PR #1, opened 2025-03-19
Add install scripts for basic implementation

dskvr:
run 

```
./scripts/install.sh
```

- installs `scp-server` 
- copies `./scripts/init.d/tollgate` to `/etc/init.d`

```
service tollgate start
service tollgate status
service tollgate restart
service tollgate stop
```


c03rad0r:
We have init.d scipts now. Thanks for pointing out that this is the right way to do it..

PR #21, opened 2025-06-20
Support Raspberry Pi

Origami74:
Tested on raspi and it works!

it accidentally included some changes for the release notes but i think it's fine to leave it here


Origami74:
Confirmed it works

PR #33, opened 2025-07-07
Feature/test joel frontend

c03rad0r:


c03rad0r:
Seems to be working, but I broke the icons and stuff when integrating it. I'll close the branch for now since the camera stuff is still work in progress

PR #37, opened 2025-07-09
Feature/identity config

c03rad0r:

PR #39, opened 2025-07-10
Simplify janitor using version number scheme

c03rad0r:


c03rad0r:
This doesn't need to be reviewed any time soon. We can just remove the janitor for now as shown here:
https://github.com/OpenTollGate/tollgate-module-basic-go/pull/38

I will wait for [this branch](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/37) to be merged before I even bother going through the manual test cases listed above, because things could change in the mean time..

PR #61, opened 2025-11-07
Free SSID for owner

Origami74:
Adds password protected SSID for owner that does NOT trigger captive portal and gets internet access immediately.

- SSID
- Random generated password, 3 words, 1 number

**CLI**
CLI tooling included to enable/disable private network, and change name/password


Origami74:
Merged manually by cobrador

PR #63, opened 2025-11-10
build opkg packages with and without upx compression (so that they work on 16 mb devices)

Amperstrand:
This compresses the go binaries with npx.

Not sure what the exact tradeoffs are, but the opkg packages can now be installed on 16 mb routers.

We might only want npx enabled on 16 mb devices and only when not as part of an actual firmware image (which has its own compression using squashfs).

This way of doing it means we have build targets that let users optionally choose the npx compressed or the normal uncompressed opkg file


c03rad0r:
Looks good, lets simplify the "none" thing like you said, test and merge :)

PR #66, opened 2025-12-10
Remove wireless interfaces from ignore list of crowsnest

Origami74:
- Also remove wildcard ignore for all bridge-interfaces `startswith(br-*)`


Origami74:
tested by @c03rad0r, looked good

PR #75, opened 2026-01-07
Data measurement

Origami74:
- uses ndsctl for downstream measurement instead of measuring the gateway
- added /usage endpoint that returns [usage]/[allotment]
- use /usage endpoint to get remaining allotment for renewals


c03rad0r (approved):
Looks good :)

PR #79, opened 2026-04-16
Add local apk package build support while preserving the opkg path

alexxie16:
## Summary
- move the package lifecycle hooks into the live OpenWrt package metadata so one `Makefile` now drives both the legacy `ipk` flow and the OpenWrt 25 `apk` flow
- derive Go target settings from the selected OpenWrt SDK target instead of relying on manually exported `GOARCH` / `GOMIPS`, so generated packages carry the correct target architecture metadata
- add local SDK verification helpers for `apk` builds and fork-based testing while keeping the on-device installer compatible with both `apk` and `opkg`

## What Changed
- `Makefile`
  - adds `Package/$(PKG_NAME)/preinst` and `Package/$(PKG_NAME)/postinst` so package hooks are generated through OpenWrt package metadata instead of by copying files into `$(1)/CONTROL`
  - removes the manual `$(1)/CONTROL` install block that conflicts with the `apk` pack path
  - introduces `TOLLGATE_DISPLAY_VERSION` and only normalizes `PKG_VERSION` when `CONFIG_USE_APK=y`, so branch-style versions can still be embedded in the binaries while the resulting `.apk` stays valid
  - derives `TOLLGATE_GOARCH` / `TOLLGATE_GOMIPS` from the SDK target architecture and uses that for both binaries instead of trusting shell-exported `GOARCH=mips`
  - adds `TOLLGATE_PKG_SOURCE_URL` so local SDK builds can verify fork-only commits without overriding source URLs for unrelated packages
- `normalize-apk-version.sh`
  - normalizes package versions into an `apk`-safe form, including fallback handling for legacy branch-style versions
- `files/usr/bin/check_package_path`
  - prefers `apk add --allow-untrusted` when `apk` is present and falls back to `opkg install` otherwise
  - captures the installer return code explicitly before clearing `update_path`
- `build-sdk-apk.sh`
  - adds a local Docker SDK helper for verbose OpenWrt 25.12 `apk` builds and artifact collection under `/tmp/tollgate-build-artifacts`
  - documents that it is a local developer helper
  - accepts `TOLLGATE_PKG_SOURCE_URL` so fork branches can be built in the SDK without editing the checked-in package recipe

## Removed
- `Makefile.legacy-opkg`
- `files/CONTROL/preinst`
- `files/CONTROL/postinst`

These were redundant once the live `Makefile` became the single source of truth for both package formats.

### Checks run
- `sh -n build-sdk-apk.sh`
- `sh -n normalize-apk-version.sh`
- `sh -n files/usr/bin/check_package_path`
- `git diff --check`

### SDK build verification
- `apk` build succeeded with OpenWrt `25.12.0` on `mediatek/filogic`:
  - artifact: `/tmp/tollgate-build-artifacts/apk-25_12/aarch64_cortex-a53/base/tollgate-wrt-0.0.0-r0.apk`
  - package metadata emitted by `apk mkpkg` shows `arch:aarch64_cortex-a53`
- legacy `ipk` build succeeded with OpenWrt `24.10.1` on `ath79/generic`:
  - artifact: `/tmp/tollgate-build-artifacts/ipk-24_10/mips_24kc/base/tollgate-wrt_fix-branch.74.e9f593d_mips_24kc.ipk`

### How to verify locally
1. Build an `apk` for a 25.12 target:
   ```bash
   SDK_TAG=mediatek-filogic-25.12.0 ./build-sdk-apk.sh
   ```
2. If you need to verify a fork-only branch in the SDK, pass the package source override:
   ```bash
   SDK_TAG=mediatek-filogic-25.12.0 \
   TOLLGATE_PKG_SOURCE_URL=https://github.com/alexxie16/tollgate-module-basic-go.git \
   ./build-sdk-apk.sh
   ```
3. Confirm the generated `apk` exists:
   ```bash
   ls -lh /tmp/tollgate-build-artifacts/apk-25_12/aarch64_cortex-a53/base/
   ```
4. Copy the package to a 25.12 router and install it:
   ```bash
   apk add --allow-untrusted /tmp/tollgate-wrt-0.0.0-r0.apk
   ```
5. Verify the router architecture if needed:
   ```bash
   apk --print-arch
   ```
6. Verify the legacy `ipk` path in a 24.10 SDK with `CONFIG_USE_APK` unset. For fork-only commits, pass `TOLLGATE_PKG_SOURCE_URL` to the package build command in the SDK environment.

## Notes
- this PR still focuses on local package build/install paths only; CI and janitor/self-update packaging changes are intentionally left out
- `apk add --allow-untrusted` matches the current local-file install behavior and is not intended to replace a future signed-feed workflow


Origami74:
I gave it an LLM code review to find the easy fixes, can you have a look at these?

---

Solid, focused change. The core idea — moving `preinst`/`postinst` from hand-copied `$(1)/CONTROL/` files into `define Package/$(PKG_NAME)/preinst|postinst` — is the correct fix for OpenWrt 25's `apk` path. Arch derivation from `$(ARCH)` instead of exported `GOARCH` is also an improvement. That said, a few things stand out:

### Concerns

**1. `Makefile.legacy-opkg` is duplication, not preservation.**
The main `Makefile` already keeps the `ipk` path working — apk-specific bits are gated on `CONFIG_USE_APK=y`, and the PR description confirms both 24.10.1 (ipk) and 25.12.0 (apk) builds succeed from the *single* Makefile. So the 181-line legacy copy will silently drift. Remove it; the gate is the whole point.

**2. `files/CONTROL/preinst` + `postinst` are now orphaned.**
With the `$(1)/CONTROL` install block removed, nothing in the live `Makefile` references them — only `Makefile.legacy-opkg` does. If you drop the legacy Makefile (point 1), also delete these files. The hook logic now lives inline in the `define Package/.../preinst|postinst` blocks, which is the source of truth. Keeping two copies is how they go out of sync.

**3. `Makefile.legacy-opkg` ships a placeholder maintainer.**
`PKG_MAINTAINER:=Your Name <your@email.com>`. If it stays, fix it; better still, drop the file.

**4. `check_package_path` exit-code check is brittle.**
The \`if [ \$? -eq 0 ]\` after the \`if/elif/else\` works today, but only because each branch's last command is the install (or an \`exit 1\`). One future edit adding a trailing \`echo\` inside a branch silently breaks the clear-on-success semantics. Capture it explicitly:
\`\`\`sh
if command -v apk >/dev/null 2>&1; then
    apk add --allow-untrusted "\$update_path"; rc=\$?
elif command -v opkg >/dev/null 2>&1; then
    opkg install "\$update_path"; rc=\$?
else
    echo "No supported package manager found" >&2; exit 1
fi
if [ "\$rc" -eq 0 ]; then ...
\`\`\`

**5. \`apk add --allow-untrusted\` skips signature verification.**
Matches the pre-existing \`opkg install\` behavior for local files, so not a regression — just flagging it in case a signed-feed path is on the roadmap.

**6. Minor — ARM32 \`GOARM\` isn't set.**
The arch mapping handles \`aarch64\`→\`arm64\` and mips variants, but OpenWrt targets like \`arm_cortex-a7_neon-vfpv4\` will just give Go \`GOARCH=arm\` with no \`GOARM\`, which defaults to \`5\`. Not caused by this PR (the old code had the same gap), but since you're reworking arch handling it'd be a cheap win: derive \`GOARM\` from the target in the same style as \`GOMIPS\`.

**7. \`normalize-apk-version.sh\`'s \`[0-9]*\` branch.**
Looks right for \`1.2.3\`, \`0.0.1-rc1\`, etc. Worth confirming behavior on \`1.2.3+meta\` (apk's \`+\` is legal but worth a manual check); not blocking.

### Nits

- \`build-sdk-apk.sh\`: installing \`golang-go\` from Debian backports inside the SDK works, but ties builds to a moving target. Fine for a local dev helper, just worth a comment at the top that this is a developer convenience, not a reproducible build.
- \`build-sdk-apk.sh\`: \`rm -rf /artifacts/apk-25_12\` clobbers prior artifacts on the host (via the bind mount). That's probably intentional, but it's easy to miss on first read.
- The preinst/postinst whitespace churn in \`files/CONTROL/*\` (spaces→tabs) is unrelated to the apk work and adds diff noise — not a blocker.

alexxie16:
> I gave it an LLM code review to find the easy fixes, can you have a look at these?
> 
> Solid, focused change. The core idea — moving `preinst`/`postinst` from hand-copied `$(1)/CONTROL/` files into `define Package/$(PKG_NAME)/preinst|postinst` — is the correct fix for OpenWrt 25's `apk` path. Arch derivation from `$(ARCH)` instead of exported `GOARCH` is also an improvement. That said, a few things stand out:
> 
> ### Concerns
> **1. `Makefile.legacy-opkg` is duplication, not preservation.** The main `Makefile` already keeps the `ipk` path working — apk-specific bits are gated on `CONFIG_USE_APK=y`, and the PR description confirms both 24.10.1 (ipk) and 25.12.0 (apk) builds succeed from the _single_ Makefile. So the 181-line legacy copy will silently drift. Remove it; the gate is the whole point.
> 
> **2. `files/CONTROL/preinst` + `postinst` are now orphaned.** With the `$(1)/CONTROL` install block removed, nothing in the live `Makefile` references them — only `Makefile.legacy-opkg` does. If you drop the legacy Makefile (point 1), also delete these files. The hook logic now lives inline in the `define Package/.../preinst|postinst` blocks, which is the source of truth. Keeping two copies is how they go out of sync.
> 
> **3. `Makefile.legacy-opkg` ships a placeholder maintainer.** `PKG_MAINTAINER:=Your Name <your@email.com>`. If it stays, fix it; better still, drop the file.
> 
> **4. `check_package_path` exit-code check is brittle.** The `if [ $? -eq 0 ]` after the `if/elif/else` works today, but only because each branch's last command is the install (or an `exit 1`). One future edit adding a trailing `echo` inside a branch silently breaks the clear-on-success semantics. Capture it explicitly: ```sh if command -v apk >/dev/null 2>&1; then apk add --allow-untrusted "$update_path"; rc=$? elif command -v opkg >/dev/null 2>&1; then opkg install "$update_path"; rc=$? else echo "No supported package manager found" >&2; exit 1 fi if [ "$rc" -eq 0 ]; then ... ```
> 
> **5. `apk add --allow-untrusted` skips signature verification.** Matches the pre-existing `opkg install` behavior for local files, so not a regression — just flagging it in case a signed-feed path is on the roadmap.
> 
> **6. Minor — ARM32 `GOARM` isn't set.** The arch mapping handles `aarch64`→`arm64` and mips variants, but OpenWrt targets like `arm_cortex-a7_neon-vfpv4` will just give Go `GOARCH=arm` with no `GOARM`, which defaults to `5`. Not caused by this PR (the old code had the same gap), but since you're reworking arch handling it'd be a cheap win: derive `GOARM` from the target in the same style as `GOMIPS`.
> 
> **7. `normalize-apk-version.sh`'s `[0-9]*` branch.** Looks right for `1.2.3`, `0.0.1-rc1`, etc. Worth confirming behavior on `1.2.3+meta` (apk's `+` is legal but worth a manual check); not blocking.
> 
> ### Nits
> * `build-sdk-apk.sh`: installing `golang-go` from Debian backports inside the SDK works, but ties builds to a moving target. Fine for a local dev helper, just worth a comment at the top that this is a developer convenience, not a reproducible build.
> * `build-sdk-apk.sh`: `rm -rf /artifacts/apk-25_12` clobbers prior artifacts on the host (via the bind mount). That's probably intentional, but it's easy to miss on first read.
> * The preinst/postinst whitespace churn in `files/CONTROL/*` (spaces→tabs) is unrelated to the apk work and adds diff noise — not a blocker.

Thanks, this is very heplful, I probably could have used another AI agent / model to review the code myself. 

I agree the main cleanup items are:
- `Makefile.legacy-opkg` is redundant if the live `Makefile` already preserves the `ipk` path behind `CONFIG_USE_APK`
- `files/CONTROL/preinst` and `postinst` should be removed if the `Package/.../preinst|postinst` blocks are now the source of truth
- `check_package_path` should capture the install return code explicitly instead of relying on `$?` after the conditional
I’m treating the `GOARM` point as a reasonable follow-up rather than expanding scope here, unless we decide to cover/test ARM32 in this PR as well.

I’m less inclined to expand scope for this PR on `GOARM` unless we want to explicitly support/test ARM32 in the same change, but I agree it’s a valid follow-up. Same for the `+meta` version case and signed package verification: good notes, but not blockers for the local apk transition.


Working on them now.

Origami74:
Agree with your points!

I tested it here as well and works great!. Also tested on OpenWRT 24 to validate the changes didn't break anything there.

It all seems to work fine, so I'll merge this now

Thank you!

PR #81, opened 2026-04-20
chore: dead-code + docs cleanup sweep

Origami74:
## Summary

Twenty focused commits removing dead code, stale docs, and developer-experience papercuts. **+201 / −5,670 net lines.** Each commit is independently revertable.

### Dead packages and helpers
- **janitor** — `initJanitor()` was already commented out in `main.go`; the package had no live caller.
- **relay (in-process Khatru server)** — `publishLocal` had zero callers, so the local NIP-01 store was permanently empty and `getLatestSession` always returned nil. With it, the `Allow-Relay-In` firewall rule on port 4242 and the `relay_in` UCI block.
- **Public-relay infrastructure** — `merchant.publishPublic` had zero callers; dropping `PublicPool`, `GetPublicPool`, the `Relays` config field, and the default relay list. Fields removed from JSON unmarshal silently — no version bump needed.
- **10 dead `config_manager` helpers** — NIP-94 fetch flow, package-info parser, install-version probe, architecture detection, `CalculateMinPayment`, `generatePrivateKey` + already-deprecated `setUsername`. All orphaned by the janitor removal.
- **`tools/generate_key.go`** — standalone main with no callers (`nak key generate` does the same).
- **`golang.mk`** — 2018 OpenWrt copy, never `include`d, `GO_PKG` never set.
- **Orphan root-of-`src/` tests** — `handlers_test.go` and `wallet_test.go` failed to even compile in package init without `/etc/tollgate/config.json`.
- **Obsolete migrations** — `98-…v0.0.1→v0.0.2` and `99-…v0.0.2→v0.0.3` UCI-defaults; current `config_version` is `v0.0.7`.
- **`known_networks.json`** — installed to `/etc/tollgate/` with placeholder content, never read by any Go code.

### Stale docs
- `docs/crowsnest.md` (module renamed to `upstream_detector`).
- `docs/work/reliability-fixes.md` (proposals already landed).
- `src/config_manager/{HLDD,LLDD,manual_test_edge_cases,unit_test_edge_cases}.md` — all described work that has shipped.
- `docs/upstream-gateway-flow.md` — folded into `docs/upstream_session_manager.md` as a new "Cross-component flow" section.
- `crowsnest` → `upstream_detector` text rename across the three remaining architecture docs (45 instances).

### Reorg
[...]

PR #82, opened 2026-04-20
fix(scripts/build-sdk-apk): forward host SHA and preserve source-URL default

Origami74:
## Summary

The `scripts/build-sdk-apk.sh` helper that ships in [#81](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/81) (and predates it) doesn't actually work — running it as documented produces no apk and exits with a confusing source-fetch error. Two stacked flaws:

### 1. `PKG_SOURCE_VERSION` was never forwarded
The Makefile resolves it via `$(shell git rev-parse HEAD)` at parse time. Inside the SDK container, that runs with `cwd=/builder` — which isn't a git repo — so it returns an empty string. The SDK then writes a zero-byte source tarball at `/builder/dl/tollgate-wrt-<version>.tar.zst`, `Build/Prepare` can't extract it, and you get:
```
zstd: can't stat /builder/dl/tollgate-wrt-0.0.0-r0.tar.zst : No such file or directory -- ignored
/builder/staging_dir/host/bin/tar: This does not look like a tar archive
ERROR: package/tollgate-wrt failed to build.
```

Fix: resolve the SHA on the host with `git -C $REPO_ROOT rev-parse HEAD`, embed it as a make command-line override (`PKG_SOURCE_VERSION=$sha`). Switch the inner `sh -lc` to double quotes so the SHA is a literal at host expansion time — make's CLI override only takes effect at parse time, before env-var substitution inside the container would matter.

### 2. Empty `TOLLGATE_PKG_SOURCE_URL` defeated the Makefile default
The script always passed `-e TOLLGATE_PKG_SOURCE_URL=""` to docker. The Makefile has `TOLLGATE_PKG_SOURCE_URL?=https://github.com/OpenTollGate/tollgate-module-basic-go.git`, but `?=` only substitutes a default when the variable is **unset** — an empty value still counts as set, so `PKG_SOURCE_URL` ended up blank and the SDK had nothing to clone.

Fix: only forward `TOLLGATE_PKG_SOURCE_URL` when the caller actually set it.

## Test plan
- [x] On the cleanup branch (sha `0286d2a`):
  ```
  SDK_TAG=mediatek-filogic-25.12.0 ARTIFACT_DIR=/tmp/foo bash scripts/build-sdk-apk.sh
  ```
  produces `/tmp/foo/apk-25_12/aarch64_cortex-a53/base/tollgate-wrt-0.0.0-r0.apk` (6.1 MB).
- [ ] Default invocation `bash scripts/build-sdk-apk.sh` (mips_24kc target) — not exercised here, but uses the same code paths.
- [ ] Override `TOLLGATE_PKG_SOURCE_URL=...` to build from a fork — not exercised, follows the same conditional.

## Notes
Branched off `cleanup/dead-code-and-docs` because the script was moved into `scripts/` there. Targets that branch as the base; will rebase if [#81](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/81) is squashed.

### DIFF
--- a/scripts/build-sdk-apk.sh
+++ b/scripts/build-sdk-apk.sh
@@ -13,28 +13,50 @@ mkdir -p "$ARTIFACT_DIR"
 # Anchor to repo root so the docker mount is correct regardless of CWD.
 REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
 
+# Resolve the source SHA on the host, where git works. The Makefile's
+# $(shell git rev-parse HEAD) runs at parse time inside the SDK container
+# with cwd=/builder (not a git repo), so it returns empty and the SDK
+# silently writes a zero-byte source tarball. Override on the make command
+# line so the SDK clones the right commit from TOLLGATE_PKG_SOURCE_URL.
+if ! PKG_SOURCE_VERSION="$(git -C "$REPO_ROOT" rev-parse HEAD 2>/dev/null)"; then
+    printf 'error: %s is not a git checkout — cannot determine PKG_SOURCE_VERSION\n' "$REPO_ROOT" >&2
+    exit 1
+fi
+
 printf 'Using SDK_TAG=%s\n' "$SDK_TAG"
 printf 'Using PACKAGE_VERSION=%s\n' "$PACKAGE_VERSION"
+printf 'Using PKG_SOURCE_VERSION=%s\n' "$PKG_SOURCE_VERSION"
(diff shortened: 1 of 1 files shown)


Origami74:
@alexxie16 can you test if this script still works for you?

alexxie16:
> @alexxie16 can you test if this script still works for you?

Yes, it works for me, though there are some overlaps fixing in my lighting support branch, I wanted to move it here but running short of time. I guess we can just go with your changes for now, I will see what I can do in lightning support PR.

PR #83, opened 2026-04-20
fix(first-boot): eliminate reboot race; speed up uci-defaults

Origami74:
## Summary

First-boot fixes for routers flashed with this package taking a long time to become reachable (and sometimes looking like they were restarting repeatedly).

- **Eliminate the reboot race in `95-random-lan-ip`.** The script scheduled `(sleep 5 && reboot) &`, but `99a`/`99b` run after it and can exceed 5s on low-end routers. OpenWrt only deletes a uci-defaults script after `exit 0`, so a killed `99a`/`99b` re-ran on the next boot, reconfiguring things twice. Fix: drop the reboot — the committed LAN IP is applied when procd brings network up.
- **Increase LAN IP entropy.** `RANDOM=$$$(date +%s)` seeds ash's LCG from PID + wall clock, but on first boot there's no RTC/NTP, so `date +%s` ≈ firmware build time → near-identical "random" subnets across devices. Each octet now reads from `/dev/urandom` via `hexdump`.
- **Drop redundant service restarts.** uci-defaults run before procd starts services, so `/etc/init.d/nodogsplash restart` in the symlink script and `service network reload` in `99b` were no-ops or fought the normal boot sequence. Removed both.
- **Merge `99a` + `99b` → `99-tollgate-setup`.** Function-based layout (`setup_uhttpd`, `setup_public_wifi`, `setup_private_network`, `commit_all`, …). Replaced the `uci_safe_set` wrapper (2–3 forks per key) with plain `uci set` for core OpenWrt configs; guards kept only where sections are genuinely optional (nodogsplash). Single commit pass at the end. Password generator uses one `/dev/urandom` read + awk instead of four `hexdump` forks and five `echo|cut|tr` pipelines.
- **Hotplug connectivity check.** Replaced unconditional `sleep 5; ping -W 5 8.8.8.8` with a poll loop over `ubus call network.interface.\$iface status`. Works when ICMP is blocked, returns as soon as the interface is actually up.
- **Makefile cleanup.** Collapsed duplicate \`\$(INSTALL_DIR) /etc/uci-defaults\` entries; reordered to lex boot order (90, 95, 99).

## Test plan

- [ ] Build the package from this branch with the OpenWrt SDK.
- [ ] Flash a router with a fresh image; time to SSH availability should be noticeably shorter than before.
- [ ] Verify only a single reboot occurs during first boot (previously 1–2 extra).
- [ ] Verify `/tmp/tollgate-setup.log` shows all setup steps completing.
- [ ] Flash two routers from the same image; confirm they land on different random LAN subnets (previously often identical).
- [ ] Verify public SSID `TollGate-XXXX-2.4GHz/-5GHz` and private SSID `c03rad0r-XXXX` both come up with matching suffix.
- [ ] Verify nodogsplash serves the captive portal from the symlinked htdocs.
- [ ] Trigger a WAN ifup after boot; confirm tollgate-wrt restarts once and the `/tmp/tollgate_initial_restart_done` flag is set.


Origami74:
Superseded by a new PR with the cleanup squashed into a single commit.

PR #87, opened 2026-04-20
fix(first-boot): sync clock before starting wallet on fresh flash

### LINKED ISSUE #42: Wallet doesn't start when time isn't correct on router - this happens after flashing image
Can set to browser time / ntp server in luci (system/system/time)

TODO: Automate this with a script that triggers on first connection to a gateway.

Origami74:
## Summary

- Adds `94-tollgate-timesync` hotplug.d/iface script that runs before `95-tollgate-restart` on WAN ifup
- Detects a stale clock by comparing `now` against the firmware build mtime (`/rom/etc/banner`, with `/etc/banner` fallback)
- Forces a blocking `ntpd -q` against UCI-configured NTP servers (fallback: `pool.ntp.org`) so `tollgate-wrt` restarts with a correct clock

## Why

After a fresh flash, `sysfixtime` sets the clock to the firmware build date and stays there until `sysntpd`'s async catch-up runs. That window is long enough for `tollgate-wrt` to start with a stale clock, at which point Cashu token validation fails and the wallet never comes up. See #42.

## Test plan

- [ ] Flash a fresh image onto a router with a build date more than a day old, confirm wallet starts on first WAN ifup without manual time setting
- [ ] Reboot a running router (non-fresh): confirm script exits early via the build-date guard and does not block hotplug
- [ ] Disconnect WAN mid-sync and reconnect: confirm flag file resets on reboot (tmpfs) but not on ifup cycles

Closes #42

### DIFF
--- a/files/etc/hotplug.d/iface/94-tollgate-timesync
+++ b/files/etc/hotplug.d/iface/94-tollgate-timesync
@@ -1,0 +1,68 @@
+#!/bin/sh
+
+# After a fresh flash the router's clock is set to the firmware build date
+# (by sysfixtime) and stays there until NTP syncs. The wallet can't validate
+# Cashu tokens against a clock that's weeks behind real time, so tollgate-wrt
+# fails to start. We force a blocking NTP sync here, before 95-tollgate-restart
+# runs, so the service comes up with a sane clock on the first WAN ifup.
+
+[ "$ACTION" = "ifup" ] || exit 0
+
+WAN_DEVICE=$(uci -q get network.wan.device)
+if [ -n "$WAN_DEVICE" ] && [ "$DEVICE" = "$WAN_DEVICE" ]; then
+    :
+elif [ "$INTERFACE" = "wan" ] || [ "$INTERFACE" = "wwan" ]; then
+    :
+else
+    exit 0
+fi
+
+FLAG_FILE="/tmp/tollgate_timesync_done"
+[ -f "$FLAG_FILE" ] && exit 0
+
+# OpenWrt's sysfixtime sets the clock to the newest mtime in /etc at boot,
+# which on a fresh flash is effectively the firmware build date. If the
+# current time is still within ~24h of that build date, NTP hasn't synced
+# yet and the clock is untrustworthy for Cashu token validation.
+#
+# /rom is the read-only squashfs; its files carry the build mtime even after
+# overlay writes change /etc timestamps. Fall back to /etc/banner if /rom
+# isn't present (e.g. in non-squashfs builds).
+BUILD_REF=/rom/etc/banner
+[ -e "$BUILD_REF" ] || BUILD_REF=/etc/banner
+
+NOW=$(date +%s)
+BUILD_TS=$(date -r "$BUILD_REF" +%s 2>/dev/null)
+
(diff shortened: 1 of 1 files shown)


c03rad0r:
**Branch:** `fix/issue-42-timesync`
**Commit:** `8a62e46` — fix(first-boot): sync clock before starting wallet on fresh flash
**Author:** @Origami74

---

Thanks for putting this together — the approach is solid and the ordering with `94-` before `95-tollgate-restart` is clean. A few items came up during review, starting with the ones worth discussing before merge.

---

## Discussion points

### Rebase onto latest main

The branch currently diverges from `efc8a92`, which is behind `origin/main`. There are no merge conflicts, but changes on `main` fix the `opkg` command used during testing, so it would be easier to verify this branch after a rebase.

### 24-hour heuristic edge case

**File:** `files/etc/hotplug.d/iface/94-tollgate-timesync:37`

If the firmware image is more than a day old at flash time, the build-date guard passes immediately even on a fresh flash with a stale clock. The wallet would still fail to start in that scenario. Curious whether this is a realistic concern for your deployment window — if images are typically flashed within a day of being built it's a non-issue, but if they sit on a shelf (or in a CI cache) for longer it might be worth tightening the check (e.g., using a persistent flag set during first-boot instead of a time-based heuristic).

### Has the happy path been tested?

The test plan in the PR description looks good — just want to confirm: has the primary scenario been verified end-to-end? (Fresh flash with a stale build-date clock, wallet comes up on first WAN ifup without manual time setting.)

---

## Minor LLM nitpicks

These are small items I noticed while reading through — nothing blocking, just flagging for awareness.

- **BusyBox `ntpd` and `-N` (`line 58`):** The `-N` flag is a GNU `ntpd` option. On OpenWrt, bare `ntpd` resolves to BusyBox's applet, which may not recognize it. Probably worth dropping to be safe.

- **Missing trailing newline at EOF (`line 68`):** The file ends without a final newline. POSIX expects one, and it can cause subtle issues with some editors and tooling.

- **Retry loop timing comment (`lines 53-54`):** The comment says "up to ~15s across 3 attempts", but BusyBox `ntpd -q` blocks for roughly 10s internally per call, making the worst case closer to ~30s. Might be worth updating the comment or reducing to a single attempt.

- **No NTP dependency in Makefile:** The `DEPENDS` line doesn't list any NTP package. BusyBox `ntpd` is present in most OpenWrt base installs but isn't guaranteed across all target configs. Could be worth adding `+busybox` or documenting the assumption.

- **UCI list parsing (`line 45`):** `uci get system.ntp.server` may return list items differently depending on the OpenWrt version (single space-separated line vs. one-per-line). The current `tr ' ' '\n'` handles the former, and the latter passes through fine — just hasn't been verified across versions.

- **Style consistency with `95-tollgate-restart`:** The companion script uses `2>/dev/null` on `uci` calls and includes debug logging, while this script uses `uci -q` and has no debug output. Not a problem now, but worth aligning on a convention for future maintainability.

---

Happy to discuss if needed. We could also just merge it if testing is difficult and we are sure that it doesn't break anything..

Origami74:
Closing — decided this belongs in tollgate-os instead, since a package-level fix doesn't help users who install tollgate-wrt on a pre-existing system. Moved to [OpenTollGate/tollgate-os#18](https://github.com/OpenTollGate/tollgate-os/pull/18).

PR #90, opened 2026-04-21
fix: execute UCI defaults and reload services in postinst

c03rad0r:
### Summary

- **postinst now applies full UCI configuration without a reboot.** The old postinst only restarted `tollgate-wrt` and manually managed the NoDogSplash symlink. UCI defaults scripts (`90-tollgate-captive-portal-symlink`, `95-random-lan-ip`, `99-tollgate-setup`) were never executed during `opkg install`, so wireless, network, firewall, and DNS settings were only applied on first boot via `/etc/init.d/boot`. Installing on an already-running router left it unconfigured until a reboot.
- **`network restart` replaces `network reload` with bridge-readiness polling.** `network reload` does not create new bridge devices (e.g. `br-private`) — it only re-applies config to existing interfaces. A full `network restart` tears down and recreates the stack. A `wait_for_iface` loop polls `/sys/class/net` for each bridge (up to 15 s) before running `wifi reload`, since wifi-ifaces that reference a non-existent bridge will silently fail to attach.
- **`99-tollgate-setup` preserves active STA connections on upgrade.** Three-layer defense: (1) a self-guard flag (`/etc/tollgate-setup-done`) prevents re-execution after first successful run; (2) `configure_radio_ap()` skips any `default_radioX` interface that is currently `mode=sta`, preserving the upstream connection; (3) `mode=ap` and `network=lan` are set explicitly so the interface can never be left in an indeterminate state.
- **`tollgate-wrt` uses `stop` + `start` instead of `restart` on first install.** On a fresh install the service was never running, so `stop` is a safe no-op and `start` ensures a clean first launch.

### Network blip during install

`/etc/init.d/network restart` tears down and rebuilds the entire network stack, which briefly drops the active STA upstream connection — even on the happy path where the STA is preserved post-restart. Operators upgrading live routers should expect a short connectivity blip (typically 2-5 s) rather than a seamless swap. The STA reconnects automatically via `wifi reload`.

### Changes

| File | Description |
|---|---|
| `Makefile` (postinst) | Run UCI defaults in order; `network restart` + `wait_for_iface` bridges before `wifi reload`; `stop`+`start` for tollgate-wrt; log hint when `br-private` times out |
| `files/etc/uci-defaults/99-tollgate-setup` | Add self-guard flag; extract `configure_radio_ap()` that skips STA interfaces; set `mode=ap` and `network=lan` explicitly; create flag on success |
| `tests/docs/throughput-measurement.md` | Removed — stale working notes, not documentation |

### Tested on

- Hardware: GL.iNet GL-MT3000 (MT7981, Filogic 830)
- OpenWrt: snapshot with mt76 wireless drivers
- Install path: `opkg install` on a running router (no reboot required)
- Upgrade path: `opkg install` over an existing installation with active STA upstream connection — STA preserved, internet remained available after install


Origami74:
Core fix is solid — postinst running uci-defaults closes a real gap, and the `configure_radio_ap` / STA-preservation refactor in `99-tollgate-setup` is clean. A few requests before merge:

### 1. Please split the `c03rad0r` → `c08r4d0r` rename out of this PR

Around 9% of this diff (11 lines across 6 files) is unrelated string renames — the private SSID in `99-tollgate-setup`, plus password/SSID constants in four test files and one doc. None of it is needed for the postinst fix, and bundling it forces reviewers to reason about two unrelated changes at once. As a general ask for future PRs: please keep cosmetic/rename sweeps in their own PR — it keeps diffs scannable and makes reverts safer if one change turns out to be wrong.

### 2. Call out the `network restart` blip in the PR body

A full `/etc/init.d/network restart` tears down and rebuilds the stack, which will briefly drop the active STA upstream during install — even on the "happy path" where the STA is preserved post-restart. The PR body says "internet remained available after install," which is true end-to-end but glosses over the transient. Worth a line in the body / release notes so operators upgrading live routers know to expect a short blip rather than a seamless swap.

### 3. `wait_for_iface br-private` fallback

On devices where the private bridge isn't configured (single-radio hardware, or a variant where `setup_private_network` bails), this will burn the full 15 s every install. Not a blocker, but a quick `log "skipping — no private bridge configured"` after the warn would save future-you some head-scratching in `/tmp/tollgate-setup.log`.

### 4. `tests/docs/throughput-measurement.md`

This file is a stale working doc — it should just be deleted rather than edited. Happy to take that as a follow-up if you'd prefer, but please don't carry the rename edit on it into this PR.

c03rad0r:
Thanks for the thorough review — all four points addressed:

### 1. `c03rad0r` → `c08r4d0r` rename — split out

Done. The rename has been reverted from this branch and moved to a standalone PR on [`refactor/rename-c03rad0r-to-c08r4d0r`](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/new/refactor/rename-c03rad0r-to-c08r4d0r). This PR's diff now contains zero rename changes.

### 2. Network restart blip — documented

Added a "Network blip during install" section to the PR body calling out that operators should expect a 2-5 s connectivity blip during `network restart`, even on the happy path.

### 3. `wait_for_iface br-private` fallback — log hint added

The `br-private` wait now prints a note on timeout:

```
Note: br-private not found — this is expected on devices without a private network configured.
```

### 4. `tests/docs/throughput-measurement.md` — deleted

Removed in [`7df0bf1`](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/90/commits/7df0bf1). The file was a stale working note with raw terminal output and open questions, not documentation.

### Diff summary

The PR now touches only the files directly related to the fix:

| File | Change |
|---|---|
| `Makefile` (postinst) | UCI defaults execution, `network restart` + bridge polling, `stop`+`start` for tollgate-wrt, br-private log hint |
| `files/etc/uci-defaults/99-tollgate-setup` | Self-guard flag, `configure_radio_ap()` STA skip, explicit `mode=ap`/`network=lan` |
| `tests/docs/throughput-measurement.md` | Deleted |

@Origami74

Origami74:
Wonderful, thanks!

PR #106, opened 2026-04-28
fix: evict expired timed sessions and start the scan loop

alexxie16:
## Summary
- expire millisecond-based sessions during `GetSession` lookups so stale in-memory sessions do not continue to report usage after their allotment is spent
- start the background connectivity scan loop before blocking on `ListenAndServe`, and drop the stale inline import comments left behind in `main.go`
- remove the unused `mintConfig` parameters from the allotment helpers and cover the session-expiry behavior with a focused merchant test

## Test Plan
- `go test ./...` (from `src/`)
- `go test github.com/OpenTollGate/tollgate-module-basic-go/src/merchant` (from `src/`)


Origami74:
Suggest sequencing this after #109 — that PR replaces the entire post-`ListenAndServe` scan goroutine with the new `UpstreamManager` daemon, so point 2 here becomes moot once #109 lands.

Once #109 is in, could you rebase this down to just points 1 and 3? The session-eviction fix in `merchant.GetSession` and the `mintConfig` parameter cleanup are independent and worth landing on their own — they shouldn't be held up by the scan-loop work.

alexxie16:
> Suggest sequencing this after #109 — that PR replaces the entire post-`ListenAndServe` scan goroutine with the new `UpstreamManager` daemon, so point 2 here becomes moot once #109 lands.
> 
> Once #109 is in, could you rebase this down to just points 1 and 3? The session-eviction fix in `merchant.GetSession` and the `mintConfig` parameter cleanup are independent and worth landing on their own — they shouldn't be held up by the scan-loop work.

Sounds good,  will do.

Amperstrand:
## Cloud Lab Test Results (Amperstrand Fork)

Tested the session expiry portion of this PR using an Amperstrand fork branch [`test/pr106-session-eviction`](https://github.com/Amperstrand/tollgate-module-basic-go/tree/test/pr106-session-eviction) (commit `c3daff8`).

### What was tested
Cherry-picked the merchant.go session expiry fix (removing unused `mintConfig` parameter from `calculateAllotmentMs`/`calculateAllotmentBytes`). The scan loop portion depends on the unmerged `gatewayManager` from upstream_wifi and was excluded.

### Results
**70 passed, 0 failed** ✅

| Runner | Passed | Failed | Skipped |
|--------|--------|--------|---------|
| API    | 69     | 0      | 143     |
| Visual | 1      | 0      | 2       |

### Test file
[`tests/api/test_session_expiry_and_scan.py`](https://github.com/OpenTollGate/physical-router-test-automation/blob/main/tests/api/test_session_expiry_and_scan.py) — 5 tests, gated with `gate_bug_fix()`.

Full report: https://tests.tollgate.me/reports/c3daff8/20260524T201357Z-test-pr/report/index.html

### Recommendation
The session expiry fix in merchant.go is verified working. The scan loop + gatewayManager changes should be tested separately once the upstream_wifi module merges.

c03rad0r:
Apologies for the late review — thank you for the patience.

## Status: Needs Rebase

This branch has merge conflicts with current `main`. Please rebase onto the latest `main` (which now includes PR #120 mint health tracking, PR #122 WGM improvements, and PR #123 SSL rewrite).

## Review of Changes

### Critical Bug Fix

**Unreachable isOnline goroutine** (`main.go`): `mainLogger.Fatal(server.ListenAndServe())` was called *before* the `isOnline` goroutine, meaning the goroutine was never reached. Moving `ListenAndServe` after the goroutine fixes this — the online check now actually runs. This is a real production bug that would have prevented automatic recovery when connectivity returns.

### Cleanups

**Removed unused `mintConfig` parameter**: `calculateAllotmentMs` and `calculateAllotmentBytes` no longer take an unused `*config_manager.MintConfig` parameter. Clean interface improvement.

### Tests

**Expired session recreation test**: `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` verifies that `AddAllotment` correctly creates a fresh session after the previous one expired and was removed by `GetSession`. This covers a real edge case where a customer's session expires and they immediately pay again.

**Fixed test return order** (`tollwallet_test.go`): `err, _ := tollWallet.Receive(token)` corrected to `_, err := tollWallet.Receive(token)` — the original was silently discarding the actual result and assigning the error to the wrong variable.

### Post-Rebase Notes

`main.go` has changed significantly on main (PR #118 adds mint health tracking, merchant provider, TLS config). The isOnline goroutine move is the key fix — make sure the order is preserved after rebase.

Amperstrand:
## Code Review

Reviewed commit `9d18ac2` + merge commit `a1b8757`. Three changes, one test. Overall: **changes are correct and minimal. Approve with one question.**

---

### 1. Scan loop move: `main.go` ✅ Correct

**Before**: `server.ListenAndServe()` (blocking) → `go scanLoop()` (never reached until server exits) → `"Shutting down"` print
**After**: `go scanLoop()` → `server.ListenAndServe()` (blocking, returns on shutdown) → `mainLogger.Fatal()` propagates error

This is a real bug. The scan loop (background connectivity checker that calls `isOnline()`) was effectively dead code at runtime — it only ran after `ListenAndServe` returned, which is when the process is shutting down. Moving it before the blocking call is the right fix.

**One concern**: The scan loop calls `isOnline()` which reads network interfaces. If it starts before the HTTP server binds, could it interfere? Looking at the code: `isOnline()` is a read-only check of `net.Interfaces()` — no side effects, no dependency on the HTTP server being up. Safe to start earlier. ✅

The `"Shutting down Tollgate - Whoami"` → `mainLogger.Fatal()` change is also correct. The old `fmt.Println` was going to stdout; `mainLogger.Fatal` logs properly and exits with non-zero. Good cleanup.

### 2. Unused `mintConfig` parameter removal: `merchant.go` ✅ Correct

`calculateAllotmentMs` and `calculateAllotmentBytes` both took a `*config_manager.MintConfig` parameter that was never used in the function body. The callers in `calculateAllotment` were passing `mintConfig` for no reason. Removing the param is a clean dead-code elimination.

Verified: `mintConfig` is not referenced in either function body on main. No other callers exist (these are private methods). ✅

### 3. Test: `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` ✅ Good

This tests a specific scenario:
1. Create a session with 1000ms allotment that started 3 seconds ago (expired)
2. Call `GetSession` → expect error (session evicted from map)
3. Call `AddAllotment` with 2000ms → expect new session created
4. Verify the new session has the right allotment and a fresh start time

**Why this matters**: `GetSession` has a side effect — it deletes expired millisecond sessions from the map (line 836: `delete(m.customerSessions, macAddress)`). Without this test, there was no coverage for the case where a customer pays again after their session expired. The `AddAllotment` path for "session doesn't exist" (line 886-894 on main) was untested for the expiry-triggered-deletion case.

**Edge cases to consider for follow-up (not blocking)**:
- What if `GetSession` is called concurrently while `AddAllotment` is running? The RWMutex should handle this, but a concurrent test would be nice for confidence.
- What about byte-based sessions? `GetSession` only evicts `milliseconds` sessions. Byte-based sessions are never expired by `GetSession` — they just return the session with remaining bytes. This is correct (bytes don't decay over time), but worth a comment in the code.
- What if `Allotment` is 0? `elapsedMs >= 0` is always true, so a 0-allotment session would be immediately expired. Probably fine (a 0-allotment session means "no time purchased"), but could be an unexpected edge case.

### 4. Import comment cleanup: `main.go` ✅ Cosmetic, correct

`context // Added for context.Background()` → `context` and `net // Added for net.Interfaces()` → `net`. Clean. These comments were noise.

---

## Summary

| Change | Verdict | Risk |
|--------|---------|------|
| Scan loop before ListenAndServe | ✅ Bugfix, correct | Low — `isOnline()` has no server dependency |
| Remove unused `mintConfig` params | ✅ Cleanup, correct | None — parameter was unused |
| Session expiry test | ✅ Good coverage | None — test-only |
| Import comment cleanup | ✅ Cosmetic | None |

**One question for the author**: The `GetSession` expiry logic (which is on main, not from this PR) uses `time.Since(time.Unix(session.StartTime, 0)).Milliseconds()`. This returns `int64`, which is cast to `uint64`. If `time.Since` returns a negative value (clock skew backwards), `uint64(negative)` wraps to a huge number, which would always be `>= allotment`, causing immediate expiry. Have you considered this? Not blocking — just something to note for a follow-up.

---

**Recommendation**: Approve. The changes are minimal, correct, and well-tested. The session expiry behavior being tested was already on main but had no test coverage — this PR fills that gap. Would be good to run through the cloud test framework to validate on the full integration, since the scan loop change affects the runtime lifecycle.

Amperstrand:
## Updated Review — Merge Conflict Analysis

After attempting to merge, I found the conflicts are **not trivial**. Here's the situation:

### Scan loop fix is already addressed on main

The primary bugfix in this PR — moving the scan loop before `ListenAndServe` — is **already fixed differently on main** via PR #122 (WGM improvements). On main:

```
Line 219: upstreamManager.Start(context.Background())  // ← starts connectivity monitor
...
Line 727: mainLogger.Fatal(server.ListenAndServe())     // ← HTTP server starts after
```

The old `gatewayManager` API that this PR's scan loop references (`GetAvailableGateways`, `ConnectToGateway`) no longer exists — it was replaced by `upstreamManager` in PR #122.

### Merge conflicts

`src/main.go` has 2 conflicts:
1. **Imports**: `"net"` — needed for `isOnline()` which uses `net.Interfaces()`. Trivial to resolve.
2. **Scan loop placement**: The PR's scan loop code calls `gatewayManager.GetAvailableGateways()` which doesn't exist on main. This is the hard conflict.

`src/merchant/merchant.go` and `src/merchant/session_test.go` merge cleanly.

### What's still valuable (mergeable)

- ✅ Removing unused `mintConfig` params from `calculateAllotmentMs`/`calculateAllotmentBytes`
- ✅ Import comment cleanup (`context // Added for...` → `context`)
- ✅ `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` — good test coverage

### Suggested path forward

1. **Cherry-pick** just the `merchant.go` + `session_test.go` changes (these merge cleanly)
2. **Drop** the `main.go` scan loop change (already fixed on main via `upstreamManager.Start()`)
3. Keep `isOnline()` as a utility function if desired, or drop it (main uses `upstreamManager.checkConnectivity()` instead)

This would be a clean, conflict-free merge of the valuable parts.

Amperstrand:
## Rebase onto `main` (drop moot scan-loop fix)

Rebased this PR onto the latest `main` (`168b97e`) to resolve merge conflicts and trim scope. Here is what changed:

### Dropped
- **`src/main.go` scan-loop reorder** — The scan loop has since been replaced by the WGM rewrite on `main`, so this hunk no longer applies and would be dead code. Removed entirely.

### Kept
- **`src/merchant/merchant.go`** — Removes unused `mintConfig` parameter from `calculateAllotmentMs` and `calculateAllotmentBytes` (cleanup).
- **`src/merchant/session_test.go`** — Adds `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` (verifies that `AddAllotment` creates a fresh session after an expired one is evicted by `GetSession`). Existing tests on `main` are preserved.

### Result
- Author attribution unchanged (Alex Xie).
- Single commit, clean rebase — no merge commit.
- Force-pushed to this branch.

Amperstrand:
## Cloud lab results

**Run:** `20260606T192618Z-1720741`

| Passed | Failed | Skipped |
|--------|--------|--------|
| 121 | 4 | 159 |

[View full report](https://tests.tollgate.me/reports/1720741/20260606T192618Z-1720741/report/index.html)

Amperstrand:
## Cloud Lab Validation — 117P/4F/133S

Tested against `refactor/merchant-session-cleanups` (commit `1720741`) in GCP nested-KVM with local mints.

| Runner | Result |
|---|---|
| visual | 0P/0F/0S (gate failed — SSH permission issue) |
| api | 117P/**4F**/133S |
| vl-scenarios | 4P/0F/26S |

### All 4 failures are pre-existing / framework issues — **not caused by this PR**:

| Test | Root Cause |
|---|---|
| `test_concurrent_payments_single_token` | "proofs already spent" — known race in concurrent token spending, pre-existing |
| `test_nutshell_wallet_v1_mint_token` | Token acceptance failure — wallet/mint compat, pre-existing |
| `test_setup_nodogsplash_gatewayport` | Expects gatewayport=2050 but this branch has the old default (80). Framework fix not in this branch |
| `test_container_reaches_openwrt_gateway` | Container SSH connectivity — infrastructure issue |

### Code Review

This is a clean, low-risk PR:
- Removes unused `mintConfig` parameter from `calculateAllotmentMs` and `calculateAllotmentBytes` — pure dead-code cleanup
- Adds `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` — verifies that expired sessions are evicted during lookup and fresh sessions are created on `AddAllotment`
- The session expiry behavior is the only behavioral change — it happens in `GetSession` (which the test verifies via direct `AddAllotment` after expiry)

**Note**: The PR body mentions "start the background connectivity scan loop before `ListenAndServe`" and "drop stale inline import comments in `main.go`" — but the diff only shows `merchant.go` and `session_test.go`. Those other changes may be missing from this commit or were already merged separately.

**Verdict**: ✅ No regressions from this PR. Ready to merge.

c03rad0r:
Hi @alexxie16 — I rebased this PR onto current `main` (incl. #161) on your behalf for the v0.5.0 push, so you don't have to. Head is now `57ee9e6`, `mergeable=CLEAN`, `behind_by=0`.

**What landed:** the `calculateAllotmentMs`/`calculateAllotmentBytes` unused `mintConfig` parameter cleanup, plus your `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` regression test (which passes — the session-expiry eviction it exercises shipped via #139). The rebase applied cleanly with no manual edits; `go test ./src/merchant/` passes locally.

Nothing else was changed. Let me know if you'd like any adjustments. — @c03rad0r

Origami74:
Thank you @alexxie16!

alexxie16:
> Hi @alexxie16 — I rebased this PR onto current `main` (incl. #161) on your behalf for the v0.5.0 push, so you don't have to. Head is now `57ee9e6`, `mergeable=CLEAN`, `behind_by=0`.
> 
> **What landed:** the `calculateAllotmentMs`/`calculateAllotmentBytes` unused `mintConfig` parameter cleanup, plus your `TestAddAllotmentCreatesFreshSessionAfterExpiredSessionIsRead` regression test (which passes — the session-expiry eviction it exercises shipped via #139). The rebase applied cleanly with no manual edits; `go test ./src/merchant/` passes locally.
> 
> Nothing else was changed. Let me know if you'd like any adjustments. — @c03rad0r

Thank you for working on it, sorry I didn't get back earlier, I haven been away from coding recently, all look great, good job! 👍🏼

PR #107, opened 2026-04-28
feat(captive portal): add Lightning checkout and balance view

alexxie16:
## Summary
- add merchant-side Lightning invoice creation and polling endpoints. (Note: not persisting quotes in disk, keeping quote state in memory only).
- bind quote polling to the MAC address that requested the invoice, switch 404 detection to `merchant.ErrQuoteNotFound`, and cache mint quote state briefly to avoid hammering external mint RPCs
- add the captive-portal Lightning UI updates plus a balance page and `/balance` endpoint so paid sessions can be tracked from the browser

## Test Plan
- `go test ./...` (from `src/`)
- `go test github.com/OpenTollGate/tollgate-module-basic-go/src/merchant` (from `src/`)
- `go build github.com/OpenTollGate/tollgate-module-basic-go/src/tollwallet` (from `src/`)

## Notes
- `go test github.com/OpenTollGate/tollgate-module-basic-go/src/tollwallet` still fails on `main` because `src/tollwallet/tollwallet_test.go` currently treats the first return value from `Receive` as an `error`.


Origami74:
Tested and works 👌 

Thank you @alexxie16!!

PR #111, opened 2026-05-04
fix(security): replace math/rand with crypto/rand for password generation

Amperstrand:
## Summary

Replaces predictable `time.Now().UnixNano()`-based randomness with `crypto/rand` in `generateRandomPassword()`. The previous implementation:

- Used `time.Now().UnixNano() % len(words)` to pick words — deterministic if called in quick succession
- Inserted `time.Sleep(1ms)` between selections to work around the predictability
- Used modulo bias (`% 100`) for the numeric suffix

The new implementation uses `crypto/rand.Int` for all random selections, removing the timing dependency and modulo bias.

## Test plan

- [x] Existing password format unchanged: `Word-Word-Word-NN`
- [x] Passwords remain WPA2-compliant (8+ chars, mixed case + digits)
- [ ] Verify on-device: `tollgate network set-private-network` generates unique passwords across rapid invocations

### DIFF
--- a/src/cli/network.go
+++ b/src/cli/network.go
@@ -1,7 +1,9 @@
 package cli
 
 import (
+	crypto_rand "crypto/rand"
 	"fmt"
+	"math/big"
 	"os/exec"
 	"strings"
 	"time"
@@ -294,7 +296,6 @@ func (s *CLIServer) handlePrivateNetworkSetPassword(newPassword string) CLIRespo
 
 // generateRandomPassword generates a human-readable random password
 func generateRandomPassword() (string, error) {
-	// Use the same word list as the shell script
 	words := []string{
 		"alpha", "bravo", "charlie", "delta", "echo", "foxtrot",
 		"golf", "hotel", "india", "juliet", "kilo", "lima",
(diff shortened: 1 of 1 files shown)


c03rad0r:
# PR Review: `fix/security-crypto-rand`

**Commit:** `0024a81` — fix(security): replace math/rand with crypto/rand for password generation

## Checklist

| # | Item | Status |
|---|------|--------|
| 1 | `crypto/rand` used correctly, no `math/rand` or `time.Now()` seeding in password path | Pass |
| 2 | Password format still `Word-Word-Word-NN` (capitalized, 2-digit suffix) | Pass |
| 3 | WPA2 length (8-63 chars) — shortest is ~16 chars | Pass |
| 4 | Empty word list edge case — theoretical only, list is hardcoded with 26 words | Noted |
| 5 | No other callers use `math/rand` for security-sensitive purposes | Pass |
| 6 | Code compiles clean | Pass |
| 7 | No unnecessary bloat — 1 file changed, all changes on-topic | Pass |
| 8 | No merge conflicts with `origin/main` — fast-forward merge possible | Pass |

## Notes

- Error propagation is clean — all 4 calls to `crypto_rand.Int` are checked
- No orphaned imports — `time` is still used elsewhere in the file
- Removed `time.Sleep` hacks are correct — no longer needed with `crypto/rand`
- No test coverage exists, but the function is self-contained and straightforward

---

Looks good to me. Feel free to merge once the CI is green.

Origami74:
Nice cleanup — `crypto/rand.Int` with `big.NewInt(n)` is the right call, and dropping the 1ms sleep is overdue. Two thoughts before merging:

## Entropy is the bigger problem

With a 26-word NATO list and a fixed `Word-Word-Word-NN` format, the keyspace is `26³ × 100 ≈ 1.76M ≈ 20.7 bits`. Once an attacker captures a WPA2 handshake, that cracks offline in seconds on a single GPU. Switching the RNG from predictable to unpredictable matters much less than the keyspace being tiny — the "fix(security)" framing oversells what this PR actually buys.

Storage/compute on the router is a non-issue: `crypto/rand.Int` is microseconds, and embedding a larger wordlist via `//go:embed` lands in the binary's read-only segment for a few KB. Two reasonable list choices:

| List | Words | Bits/word | ipk delta (gzipped) | 4 words + 2 digits |
|---|---|---|---|---|
| EFF short | 1,296 | 10.3 | ~3 KB | ~48 bits |
| BIP-39 English | 2,048 | 11.0 | ~6 KB | ~50 bits |

Either gets you from "cracks in seconds" to "cracks in years" with no meaningful cost. EFF short averages ~4.5 chars/word so the password stays phone-typeable (~23 chars).

Happy for that to be a follow-up rather than blocking this PR — but worth filing.

## Nit: `crypto_rand` alias

There's no `math/rand` in this file to clash with, so a plain `import \"crypto/rand\"` and `rand.Int(rand.Reader, …)` reads more naturally.

Otherwise looks good.

PR #112, opened 2026-05-04
feat(config): schema-driven config system with dot-path setter

Amperstrand:
## Summary

Adds a schema describing every editable config field, a reflection-based `SetDotPath()` for dot-notation config updates, profit-share validation, thread-safe `ConfigManager` (`sync.RWMutex`), and comprehensive drift tests.

This is pure backend infrastructure with **zero LuCI dependency**. Any future tooling (CLI, API, validation) can build on it.

## Changes

**New files:**
- `config_schema.go` — `FieldSchema` type, `GetConfigSchema()`, `GetIdentitiesSchema()` describing all editable fields
- `config_dotpath.go` — Reflection-based `SetDotPath(cm, "accepted_mints.0.price_per_step", "5")`
- `config_schema_test.go` — Schema field coverage + JSON serialization tests (270 lines)
- `config_schema_drift_test.go` — Ensures schema stays in sync with struct definitions + dot-path round-trips (420 lines)
- `config_drift_test.go` — Ensures JSON fields match UI expectations (221 lines)

**Modified files:**
- `config_manager.go` — Added `sync.RWMutex`, `ReloadConfig()`, `ReloadIdentities()`, RLock on all getters
- `config_manager_config.go` — Added `ValidateProfitShare()`, boot-time profit_share validation in `EnsureDefaultConfig()`
- `config_manager_identities.go` — `SaveIdentities` uses `0600` perms (contains keys)
- `config_manager_test.go` — Added 8 `TestValidateProfitShare` cases

## Design decisions

- **No new dependencies** — uses stdlib `log` instead of logrus to keep the dep tree small
- **Schema is runtime data** — `GetConfigSchema()` returns Go structs, serialized to JSON on demand. The LuCI UI calls a CLI command to get it, rather than hardcoding field names in JS.
- **`duration` type stored as nanosecond int** — Go `time.Duration` is `int64`; schema declares `type: "duration"` so consumers know to format/parse accordingly.
- **Profit share validation** — factors must sum to 1.0 within `1e-6` tolerance. Invalid profit_share at boot resets to defaults (doesn't crash the service).

## Test plan

- [x] `go test ./src/config_manager/` — all tests pass
- [x] Schema covers all `Config` struct JSON tags (drift test)
- [x] Schema covers all `IdentitiesConfig` struct JSON tags (drift test)
- [x] No orphan schema entries (reverse drift test)
- [x] `SetDotPath` round-trips for every editable primitive field
- [x] `ValidateProfitShare` catches negative, >1.0, empty, and sum != 1.0

### DIFF
--- a/src/config_manager/config_dotpath.go
+++ b/src/config_manager/config_dotpath.go
@@ -1,0 +1,151 @@
+package config_manager
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func SetDotPath(cm *ConfigManager, key, value string) error {
+	cfg := cm.GetConfig()
+	identities := cm.GetIdentities()
+
+	parts := strings.Split(key, ".")
+	if len(parts) == 0 {
+		return fmt.Errorf("empty key")
+	}
+
+	root := parts[0]
+	rest := parts[1:]
+
+	switch root {
+	case "identities":
+		if len(rest) == 0 {
+			return fmt.Errorf("cannot replace entire identities object; use specific fields")
+		}
+		err := setFieldReflect(reflect.ValueOf(identities).Elem(), rest, value)
+		if err != nil {
+			return err
+		}
+		return SaveIdentities(cm.IdentitiesFilePath, identities)
+	default:
+		err := setFieldReflect(reflect.ValueOf(cfg).Elem(), parts, value)
+		if err != nil {
+			return err
+		}
+		return SaveConfig(cm.ConfigFilePath, cfg)
+	}
+}
+
+func setFieldReflect(v reflect.Value, parts []string, value string) error {
+	for i, part := range parts {
+		if !v.IsValid() {
+			return fmt.Errorf("invalid path at %s", strings.Join(parts[:i], "."))
+		}
+
+		if v.Kind() == reflect.Ptr {
+			if v.IsNil() {
+				return fmt.Errorf("nil pointer at %s", strings.Join(parts[:i], "."))
+			}
+			v = v.Elem()
+		}
+
+		isLast := i == len(parts)-1
+
+		if idx, err := strconv.Atoi(part); err == nil && v.Kind() == reflect.Slice {
+			if idx < 0 || idx >= v.Len() {
+				return fmt.Errorf("index %d out of range (len=%d) at %s", idx, v.Len(), strings.Join(parts[:i], "."))
+			}
+			v = v.Index(idx)
+			if isLast {
+				return setFieldValue(v, value)
+			}
+			continue
+		}
+
+		if v.Kind() == reflect.Struct {
+			field, found := findFieldByJSONTag(v, part)
+			if !found {
+				return fmt.Errorf("unknown field %s at %s", part, strings.Join(parts[:i+1], "."))
+			}
+			v = field
+			if isLast {
+				return setFieldValue(v, value)
+			}
+			continue
+		}
+
+		return fmt.Errorf("cannot navigate into %s at %s", v.Kind(), strings.Join(parts[:i], "."))
+	}
+
+	return fmt.Errorf("empty path")
+}
+
+func setFieldValue(v reflect.Value, value string) error {
+	if !v.CanSet() {
+		return fmt.Errorf("field is not settable")
+	}
+
+	switch v.Kind() {
+	case reflect.String:
+		v.SetString(value)
+	case reflect.Bool:
+		b, err := strconv.ParseBool(value)
+		if err != nil {
+			return fmt.Errorf("invalid bool value %q: %w", value, err)
+		}
+		v.SetBool(b)
+	case reflect.Int, reflect.Int64:
+		if v.Type() == reflect.TypeOf(time.Duration(0)) {
+			d, err := time.ParseDuration(value)
+			if err != nil {
+				return fmt.Errorf("invalid duration %q (use e.g. 10s, 2m, 300ms): %w", value, err)
+			}
+			v.SetInt(int64(d))
+			return nil
+		}
+		n, err := strconv.ParseInt(value, 10, 64)
+		if err != nil {
+			return fmt.Errorf("invalid integer value %q: %w", value, err)
+		}
+		v.SetInt(n)
+	case reflect.Uint, reflect.Uint64:
+		n, err := strconv.ParseUint(value, 10, 64)
+		if err != nil {
+			return fmt.Errorf("invalid unsigned integer value %q: %w", value, err)
+		}
+		v.SetUint(n)
+	case reflect.Float64:
+		f, err := strconv.ParseFloat(value, 64)
+		if err != nil {
+			return fmt.Errorf("invalid float value %q: %w", value, err)
+		}
+		v.SetFloat(f)
+	default:
+		return fmt.Errorf("unsupported type %s for set", v.Kind())
+	}
+
+	return nil
+}
+
+func findFieldByJSONTag(v reflect.Value, tag string) (reflect.Value, bool) {
+	t := v.Type()
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+		jsonTag := field.Tag.Get("json")
+		if jsonTag == "" {
+			continue
+		}
+		parts := strings.Split(jsonTag, ",")
+		name := parts[0]
+		if name == "" {
+			name = field.Name
+		}
+		if name == tag {
+			return v.Field(i), true
+		}
+	}
+	return reflect.Value{}, false
+}
--- a/src/config_manager/config_drift_test.go
+++ b/src/config_manager/config_drift_test.go
@@ -1,0 +1,221 @@
+package config_manager
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+// TestConfigJSONFields verifies that the Config struct produces all JSON fields
+// expected by the LuCI UI (settings.js). If this test fails, either:
+//   - A new field was added to Config but the UI doesn't render it (update settings.js)
+//   - A field was removed from Config but the UI still references it (update settings.js)
+//   - A JSON tag was renamed (update settings.js accordingly)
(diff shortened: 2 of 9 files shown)


c03rad0r:
# PR #112 Review — Config Schema & DotPath System

## Factual Accuracy of HANDOVER.md

| Claim | Status |
|---|---|
| 9 files changed, 1,365 insertions, 13 deletions | Verified exact match |
| Base `283b58f`, tip `951215e` | Confirmed |
| ~22 tests | 23 top-level test functions — accurate |
| File-by-file descriptions | All accurate |
| Standalone, no external dependencies | Confirmed |

---

## Issues Found

### 1. Thread safety bug in `SetDotPath` — **High**

**File:** `config_dotpath.go:12-14`

`cm.GetConfig()` acquires `RLock`, returns the internal `*Config` pointer, then releases the lock. `setFieldReflect` then mutates the **shared object** without holding any write lock. Concurrent readers calling `GetConfig()` can observe partially-modified state.

The same issue exists in the pre-existing `UpdatePricing()` (`config_manager.go:110-129`).

**Recommendation:** `SetDotPath` should acquire `cm.mu.Lock()` for the entire read-modify-write cycle, or use a copy-and-swap pattern:

```go
func SetDotPath(cm *ConfigManager, key, value string) error {
    cm.mu.Lock()
    defer cm.mu.Unlock()
    // ... work directly with cm.config / cm.identitiesConfig ...
}
```

---

### 2. Missing `only_interfaces` in drift test — **Medium**

**File:** `config_drift_test.go:64-72`

The `crowsnestFields` list checks 6 of 7 `UpstreamDetectorConfig` JSON fields. `only_interfaces` is omitted:

```go
crowsnestFields := []string{
    "probe_timeout",
    "probe_retry_count",
    "probe_retry_delay",
    "require_valid_signature",
    "ignore_interfaces",
    "discovery_timeout",
    // MISSING: "only_interfaces"
}
```

Both the struct and the schema include it. If a future change removes it, this test won't catch the UI breakage.

---

### 3. No schema-constraint validation in `SetDotPath` — **Low**

The schema defines `Enum`, `Min`, `Max` constraints (e.g. `margin` has `Min: 0.0, Max: 1.0`; `log_level` has `Enum: ["debug", "info", "warn", "error"]`), but `SetDotPath` ignores them entirely. For example:

```
SetDotPath(cm, "margin", "5.0")   // succeeds despite Max: 1.0
SetDotPath(cm, "log_level", "banana") // succeeds despite Enum constraint
```

This is acceptable if validation is deferred to PR #113's CLI handler, but should be documented as intentional or tracked as follow-up work.

---

### 4. Implicit pointer-sharing coupling — **Low**

For the config path, `SetDotPath` modifies the pointer returned by `GetConfig()`, which IS `cm.config` — so the in-memory state stays correct. The same is true for identities via `GetIdentities()`. Both paths work because pointers are shared, but this coupling is fragile and implicit. A future refactor to return copies would silently break in-memory consistency.

**Recommendation:** Add a comment on `GetConfig()` / `GetIdentities()` noting that the returned pointer is shared mutable state, or switch to copy-on-read if safe mutation is preferred.

---

## Documentation Suggestions for HANDOVER.md

### 5. Thread safety section understates the problem

The "What to Review" section says:

> *Thread safety — ConfigManager now has sync.RWMutex. Verify read operations use RLock() and write operations use Lock() consistently.*

In reality, the write paths (`SetDotPath`, `UpdatePricing`) don't use `Lock()` at all — they use `RLock()` then mutate the shared object. This should be called out explicitly as a known issue or the fix should be in scope for this PR.

### 6. `duration` type is non-standard

The schema uses `Type: "duration"` for fields like `probe_timeout`, `probe_retry_delay`, `data_monitoring_interval`. This isn't a standard JSON Schema type. PR #114 (LuCI frontend consumer) will need special handling for these fields — worth noting in the handover so the frontend PR knows what to expect.

### 7. Drift tests encode a LuCI contract

The handover says this PR has "zero imports or references to LuCI/frontend code." This is true for imports, but `config_drift_test.go` comments reference `settings.js` and "LuCI UI" extensively, and the test assertions encode the specific fields the LuCI UI depends on. Not a blocker, but worth noting for reviewers — these tests serve as a cross-repo contract.

---

## Summary

The handover is well-structured and all factual claims check out. The **main actionable item** is the thread safety bug in `SetDotPath` (Issue #1) — I'd recommend fixing it before merge since PRs #113 and #114 build directly on this function. Issue #2 (missing drift test field) is a quick fix. The rest are documentation or low-priority items that can be tracked separately.

Origami74:
Really like the direction here — schema in Go, drift tests against the struct, profit-share validation with a friendly "ratio not percentage" message, and the 0600 perms on `identities.json` are all wins. The reflection-based drift test is the standout: it's the right way to keep a hand-written schema honest.

Three things I'd want addressed before merge, plus a couple of nits.

## 1. The `sync.RWMutex` doesn't protect what it looks like it protects

**In laymen terms:** the lock only guards the pointer swap, not the data behind it. Callers get a raw pointer and mutate the underlying struct outside the lock, so concurrent readers can see torn state. The right fix is copy-on-write (or `atomic.Pointer`) so readers always hold an immutable snapshot.


`GetConfig()` returns `*Config` and immediately drops the RLock ([config_manager.go:65-69](src/config_manager/config_manager.go#L65-L69)). Every caller then dereferences a raw pointer with no lock held — and several mutate through it:

- [`UpdatePricing`](src/config_manager/config_manager.go#L125-L148) writes `config.AcceptedMints[0].PricePerStep` with no write lock.
- [`SetDotPath`](src/config_manager/config_dotpath.go#L11-L39) does the same via reflection.
- Long-lived readers in [merchant](src/merchant/merchant.go#L72), [upstream_session_manager](src/upstream_session_manager/session.go#L67), and [wireless_gateway_manager](src/wireless_gateway_manager/wireless_gateway_manager.go#L63) hold the pointer across calls.

`ReloadConfig` atomically swaps the pointer, but if any goroutine wrote through the *previous* pointer, readers see torn state. `-race` passes only because no test runs `SetDotPath`/`UpdatePricing` concurrently with a hot reader.

Copy-on-write is the natural fit for this design: writers build a new `Config` (deep copy → mutate → swap pointer under the write lock), readers treat the returned pointer as immutable. `atomic.Pointer[Config]` would make it lock-free for readers entirely. As written, the mutex advertises safety it doesn't deliver — arguably worse than no mutex.

## 2. `SetDotPath` ignores the schema's own constraints

**In laymen terms:** the schema declares `Min`/`Max`/`Enum` for several fields, but the setter never checks them. Out-of-range and out-of-enum values are accepted and persisted. Cheap fix: validate against the schema entry before writing.


The schema declares `Min`/`Max`/`Enum` for `Margin`, `Metric`, `LogLevel`, `ProfitShare.Factor`, `Trust.DefaultPolicy`, etc. — but `setFieldValue` validates none of them. All of these succeed today:

```
SetDotPath(cm, "metric", "foo")                  // not in enum
SetDotPath(cm, "margin", "2.5")                  // > max
SetDotPath(cm, "profit_share.0.factor", "-1.0")  // < min, also breaks sum
```

The third one combines with #3 below into a real footgun: one CLI typo writes the file, next boot silently resets `profit_share` to defaults, payouts go elsewhere.

A schema-driven setter that doesn't enforce its schema is half a feature. Even just `Enum` and `Min`/`Max` checks before `setFieldValue` would close this.

## 3. Invalid `profit_share` is silently overwritten on boot

**In laymen terms:** if `profit_share` fails validation at boot, the user's setting is overwritten with defaults and only a `log.Printf` records it. The two neighbouring branches in the same function call `backupAndLog` to preserve the bad file — this branch should do the same so the original is recoverable.


[config_manager_config.go:249-253](src/config_manager/config_manager_config.go#L249-L253):

```go
if profitShareErr != nil {
    log.Printf("WARNING: Invalid profit_share, resetting to defaults: %v", profitShareErr)
    config.ProfitShare = defaultConfig.ProfitShare
    return &config, SaveConfig(filePath, &config)
}
```

The two adjacent branches (JSON parse error, version mismatch) call `backupAndLog` to preserve the bad file. This branch should too — otherwise a one-character typo silently redirects payouts with only a log line as evidence.

## Smaller things

- **Reflection misses `int{8,16,32}` and `uint{8,16,32}`** ([config_dotpath.go:100-119](src/config_manager/config_dotpath.go#L100-L119)). Today only `int`/`uint64` fields exist so it works, but cheap to make complete now rather than after the next field is added.
- **`config_drift_test.go` (221 lines) duplicates `config_schema_drift_test.go` (420 lines)** — the reflection-based version already catches every drift the hardcoded one does. Pick one.
- **`config_schema_test.go` is mostly smoke tests** that would be impossible-to-fail if the structs compile. Worth trimming.
- **Logical follow-up**: derive schema from struct tags (`schema:"min=0,max=1,enum=bytes|milliseconds"`) rather than maintaining it separately. Eliminates drift by construction. Out of scope for this PR — file as future work.
- **Nit**: indentation at [config_schema.go:68](src/config_manager/config_schema.go#L68) is off by one tab from its siblings.

The bones are right; the locking and validation gaps are introduced by this PR's own abstractions, so it's the natural place to fix them.

c03rad0r:
@Amperstrand feel free to call me if you want to discuss..

c03rad0r:
Superseded by #124.

Cherry-picked into develop with the following fixes on top:
- Thread safety fix: `SetDotPath` now uses `cm.mu.Lock()` + direct field access instead of `GetConfig()` (RLock) before mutation
- Schema constraint validation: `validateAgainstSchema()` checks enum/min/max before any `SetDotPath` mutation
- Added `upstream_wifi` schema entries for `UpstreamWifiConfig`
- Added schema drift tests to catch missing schema entries
- Fixed production bug: `000_main_test_env.go` was compiled into the binary, redirecting all config writes to `/tmp`

All original code from this PR is present in #124 commit `bd81d94` (feat: schema-driven config system with dot-path setter) and `61317af` (fix: thread safety, schema validation, upstream_wifi schema).

PR #18, opened 2025-06-12
Session extension - ❗️ BREAKING CHANGES ❗️

Origami74:
❗️ PROTOCOL BREAKING CHANGES ❗️

This update allows customers to extend existing sessions, this is a prerequisite for router-to-router payments

Core updates
- ❗️ Implement protocol as defined in https://github.com/OpenTollGate/tollgate/pull/5 which breaks compatibility with the captive portal.
- Added local relay + open firewall for that relay `:4242`
- Publish session events
- New payment extends latest session
- http reply with Notice events when there are issues, also publishing to relay

Current TODO's:
- [ ] Update Discovery event to spec (cashu pricing + correctly set supported TIP's)
- [ ] Update captive portal to match new spec (other repo)
- [ ] Ensure deployed tollgates also immediately get the captive portal update

PR #118, opened 2026-05-07
PR: Mint Health Tracking with Graceful Degradation

c03rad0r:
# Mint Health Tracking with Graceful Degradation

## What You Get by Merging This Branch

Merging `94-mint-health-rebase-clean` into `main` gives the tollgate router the ability to **boot and operate when all Cashu mints are unreachable**, and **automatically recover when mints come back online**. It also fixes a captive portal payment hang caused by TLS incompatibility between the router and the mint.

**New capabilities:**
- Router boots into **degraded mode** instead of crashing when mints are unreachable at startup
- **Offline wallet** loads from disk (no mint communication needed) — balance queries and cached keysets work without internet
- **Automatic recovery**: when a mint comes back, the service upgrades from degraded to full merchant in-process (no restart)
- **Dynamic downgrade**: if all mints go offline mid-operation, the service drops to degraded mode automatically
- **Health probing**: periodic `GET /v1/info` probes every 5 minutes with hysteresis (3 consecutive successes required for recovery)
- **Captive portal payments work reliably**: TLS 1.2 forced for router compatibility, 30s backend timeout, 60s frontend AbortController timeout
- **TollGate-aware upstream detection**: WGM probes `gateway:2121` after every upstream switch, adjusts patience for TollGate upstreams
- **`merchant_types` decoupling**: USM imports a 4-method interface instead of the full merchant package, eliminating ~60 lines of transitive deps (bbolt, btcwallet, lnd, etc.)
- **Cross-radio DHCP fix**: dual-trigger `ifup wwan` nudge prevents 180s timeout when switching STAs across radios
- **Startup connectivity hygiene**: emergency scan+switch if non-internet STA is active after power cycle

**Tested at commit `a701c2a` on two physical GL.iNet MT3000 routers — 22/22 hardware tests pass.**

---

## PR Size Overview

After rebasing onto `main` + `fix/wgm-improvements` (WGM PR merged), the final reviewable diff breaks down as:

| Category | Files | Net Lines | Notes |
|---|---|---|---|
| **Source code** (*.go, non-test) | 18 | +1,120 / -151 | Core feature logic |
| **Test code** (*_test.go) | 11 | +4,397 / -0 | Unit + integration tests |
| **Frontend assets** (js/css/html) | 14 | ~0 net | Bundle rebuild (renames) |
| **Stray docs** (delete before merge) | 4 | +730 | Will be removed |
[...]

PR #126, opened 2026-05-21
feat: add V2 keyset ID support (CDK 0.16.0+ compatibility)

Amperstrand:
## Summary

Adds V2 keyset ID support to the Go backend by redirecting the `gonuts-tollgate` dependency to the V2 branch ([Amperstrand/gonuts-tollgate#1](https://github.com/Amperstrand/gonuts-tollgate/pull/1)).

## Problem

CDK 0.16.0+ mints generate **V2 keyset IDs** (`01`-prefix, 33 bytes, e.g. `01df97b6fb8a572a718d7df7fcbf4387e2d455134ea8004c9c8c51e1b3391f909e`). The current gonuts only supports V1 (`00`-prefix, 8 bytes). When the Go backend connects to a CDK mint, it crashes on startup:

```
FATAL: error adding new mint: Got invalid keyset. Derived id: '0016f5fb5e5278f2' but got '01df97b6...' from mint
```

This is tracked as #18.

## Changes

- Added `replace` directive in `src/tollwallet/go.mod` and `src/merchant/go.mod` pointing `github.com/Origami74/gonuts-tollgate` to `github.com/Amperstrand/gonuts-tollgate` branch `feature/v2-keyset-ids`
- No code changes — the V2 logic (`DeriveKeysetIdV2()`, `IsKeysetIdV2()`, `getKeysetMetadata()`) lives entirely in the gonuts library

## V2 Keyset ID Spec

Per [NUT-02 PR #182](https://github.com/cashubtc/nuts/pull/182) (merged Jan 2026):
- **V1**: `00` + first 14 hex chars of SHA256(concat of raw pubkeys)
- **V2**: `01` + SHA256(`amount:pubkey_hex` pairs sorted, comma-separated, `|unit:sat`)

The V2 branch is **backward compatible** — V1 keysets continue to work via the existing derivation path.

## Upstream Dependency

This PR uses a `replace` directive until [Amperstrand/gonuts-tollgate#1](https://github.com/Amperstrand/gonuts-tollgate/pull/1) is merged and tagged. Once that happens, the `replace` directives should be removed and the versioned dependency updated.

## Testing

- CI will build all architectures including x86_64
- Cloud test framework can validate against a CDK mint
- Hardware tests (mipsel_24kc) should be run to confirm no regression with V1 mints


c03rad0r (commented):
## Review: V2 Keyset ID Support

The integration approach is clean — no Go code changes in this repo; all V2 keyset logic lives in the forked `gonuts-tollgate` library. The PR is purely a dependency swap.

### What This Does

Redirects `github.com/Origami74/gonuts-tollgate` → `github.com/Amperstrand/gonuts-tollgate` via `replace` directives in `src/go.mod`, `src/merchant/go.mod`, and `src/tollwallet/go.mod`. The Amperstrand fork adds V2 keyset ID support (01-prefix, 33 bytes) needed for CDK 0.16.0+ mints.

### Concerns

**1. Dependency downgrades are extensive.** The go.sum diff shows many transitive dependencies being downgraded or changed (etcd 3.5.16→3.5.12, docker 27.3.1→27.3.0, otel packages, grpc 1.67→1.64, prometheus client, etc.). This suggests the fork is based on an older gonuts snapshot that pulls in a different dependency tree. These downgrades could reintroduce bugs or vulnerabilities fixed in newer versions. Would be good to understand why the fork diverges so much on transitive deps — is it based on a much older commit?

**2. Fork as a permanent dependency.** Using a `replace` directive to a fork works for now, but long-term this should either:
  - Be upstreamed to Origami74/gonuts-tollgate (preferred)
  - Or the fork should be tagged with a proper semver version instead of a commit hash

**3. No integration test for V2 keysets.** Since the V2 logic is entirely in the library, there's no test in this repo that verifies a V2 keyset actually works end-to-end. If the fork has a regression, we won't catch it here. Consider adding a minimal test that creates a wallet with a V2 keyset ID and verifies it doesn't crash.

**4. go.sum churn in tollwallet.** The `src/tollwallet/go.sum` has ~150 lines changed, many are downgrades (dario.cat 1.0.1→1.0.0, coreos/go-semver 0.3.1→0.3.0, pgx v5.7.1→v5.5.4, etc.). These seem unrelated to keyset ID support and may cause issues if other modules expect the newer versions.

### Summary

The V2 keyset support itself is needed — CDK 0.16.0+ mints are already in production. The dependency swap approach is the right call for now. Flagging the transitive dependency churn as something to keep an eye on, especially if it causes build conflicts with PR #118's merchant changes.

c03rad0r (approved):
Looks good to me, ready to merge. Critical for CDK 0.16.0+ mint compatibility.

c03rad0r:
## E2E Hardware Validation (2026-05-28)

Alpha: 95P/0F | Beta: 92P/0F — no regressions. All keyset ID version tests pass on both routers.

c03rad0r (approved):
E2E validated: Alpha 95P/0F, Beta 92P/0F. V2 keyset ID support confirmed.

Origami74 (approved):

Amperstrand:
## Cloud lab finding: V2 keyset IDs silently broken with CDK mints

Running the cloud test suite against `main` (commit f4dbf73), the worker logged:

```
Go backend — skipping CDK V2 (gonuts /swap serializes V2 keyset IDs wrong)
```

This causes the test suite to silently skip CDK V2 mint tests and fall back to Nutshell V1. The V2 keyset support added in this PR (#126) appears to have a serialization issue in the gonuts library that prevents CDK 0.16.0+ mints from being used correctly.

**Impact:** any operator configuring a CDK 0.16.0+ mint with V2 keysets will likely hit the same issue. The backend doesn't crash — it silently uses the wrong keyset ID format, causing payment verification failures.

**Needs investigation:**
1. Where in gonuts does the V2 keyset ID serialization go wrong?
2. Is this the same gonuts issue referenced in #134 (gonuts-tollgate transitive dep churn)?
3. Should the V2 keyset path be feature-flagged until gonuts is fixed?

This was discovered during the v0.5.0 cloud test campaign (see #169 for full results).

Amperstrand:
## Root cause analysis: V2 keyset swap serialization

Traced the full V2 keyset path through gonuts-tollgate. The derivation is correct; the swap serialization is the known gap.

### What works (v0.7.1)

**V2 keyset ID derivation** (`crypto/keyset.go`):
```go
func DeriveKeysetIdV2(keyset PublicKeys, unit string, inputFeePpk uint) string {
    // ... sha256 hash of pubkeys + unit + fee
    return "01" + hex.EncodeToString(h[:])  // → 66 hex chars = 33 bytes ✓
}
```
Produces correct NUT-02 V2 keyset IDs. Tested against known vectors.

**V2 keyset detection** (`crypto/keyset.go`):
```go
func IsKeysetIdV2(id string) bool {
    return len(id) > 2 && id[:2] == "01"
}
```

**V2 keyset loading** (`wallet/keyset.go:66-90`):
```go
func GetKeysetKeys(mintURL, id string) (crypto.PublicKeys, error) {
    if crypto.IsKeysetIdV2(id) {
        unit, inputFeePpk, _ := getKeysetMetadata(mintURL, id)
        derivedId = crypto.DeriveKeysetIdV2(keyset.Keys, unit, inputFeePpk)
    } else {
        derivedId = crypto.DeriveKeysetId(keyset.Keys)
    }
    // validates derived ID matches mint response
}
```
Correctly branches V1/V2 derivation and validates against the mint.

### What does NOT work: /swap serialization

The cloud worker documented the exact error (CDK rejects):
```
NUT02: ID length invalid, expected 8 bytes (short/v1) or 33 bytes (v2)
```

The swap request (`nut03.PostSwapRequest`) contains:
- `Inputs` (`cashu.Proofs`): each proof has `Id string` — the keyset ID of the minting keyset
- `Outputs` (`cashu.BlindedMessages`): each has `Id string` — the keyset ID of the active keyset

Both `Proof.Id` and `BlindedMessage.Id` are typed as `string json:"id"`. The V2 hex string (66 chars) is stored correctly in Go memory. **But somewhere in the JSON serialization or HTTP request path, the V2 keyset ID ends up with the wrong byte length at the CDK mint.**

### Version inconsistency (separate but related)

| Go module | gonuts replace | version |
|---|---|---|
| `src/go.mod` (root) | `OpenTollGate/gonuts-tollgate` | **v0.7.1** (has V2 derivation) |
| `src/tollwallet/go.mod` | `Amperstrand/gonuts-tollgate` | **v0.7.0** (NO V2 derivation) |
| `src/merchant/go.mod` | `Amperstrand/gonuts-tollgate` | **v0.7.0** (NO V2 derivation) |

The root workspace build uses v0.7.1 (correct). But standalone module builds use v0.7.0 which lacks V2 support entirely. This is the same class of bug as #172 (missing replace directives).

The Amperstrand fork only has tags up to v0.7.0. The v0.7.1 tag exists only on the OpenTollGate fork. Someone needs to either:
1. Tag Amperstrand/gonuts-tollgate main as v0.7.1 (since PR #1 was merged there), OR
2. Switch all module replace directives to use `OpenTollGate/gonuts-tollgate v0.7.1`

### What needs to happen

1. **Debug the swap serialization**: create a minimal Go test that performs a `/swap` against a CDK V2 mint and dumps the raw HTTP request body. Compare the keyset ID length in the JSON to what CDK expects. The bug is likely in how `cashu.Proofs` or `cashu.BlindedMessages` serialize the `Id` field — possibly a custom `MarshalJSON` that truncates, or a proof-construction path that copies the wrong ID.

2. **Fix the version inconsistency**: align all standalone module replace directives to `OpenTollGate/gonuts-tollgate v0.7.1` (or tag Amperstrand main as v0.7.1).

3. **Add a V2 swap regression test**: once fixed, the cloud worker's "skip CDK V2" guard in `lib/cloud_lab/worker/mints.py:375` should be removed and a V2 swap test added to prevent regression.

### Is this a v0.5.0 blocker?

**Only if CDK 0.16.0+ mint support is a release goal.** If v0.5.0 targets Nutshell mints only (V1 keysets), this can ship as-is with a documented limitation. If CDK V2 mint compatibility is required, the swap serialization bug must be fixed first.

The current default dev-split (#165) and the tag-readiness funded autopay test both use Nutshell V1 mints, which work correctly. CDK V2 is not blocking any verified flow — it's a forward-compatibility gap.

PR #162, opened 2026-06-06
fix: division-by-zero, payout underflow, stale timer race

Amperstrand:
## Why This Matters

Three bugs found during pre-release audit. Two can cause **free internet**; one crashes the payment system on a common misconfiguration.

---

## Bug 1: Division-by-zero panic in `calculateAllotment` — 🔴 CRASH

**What happens**: If `price_per_step` is 0 in config (misconfigured, corrupted, or defaulted incorrectly), every payment attempt causes a **runtime panic** that crashes the merchant. The entire payment system goes down — no one can pay for internet.

**How to trigger**: Set `price_per_step: 0` in config. Every single payment attempt panics.

**Fix** (commit `165af49`): Runtime guard returns error before division. Schema-level `Min: uint64(1)` prevents 0 via config setters (defense in depth).

**Files**: `src/merchant/merchant.go`, `src/config_manager/config_schema.go`

---

## Bug 2: uint64 underflow in `processPayout` — 🔴 BILLING ERROR

**What happens**: `balance - MinBalance` wraps to ~18 quintillion sats when `balance < MinBalance`. The merchant attempts to pay out an absurdly large amount via Lightning. Whether this succeeds depends on wallet balance, but the intent is clearly wrong.

**How to trigger**: Set `min_payout_amount` lower than `min_balance` in config. These are independently configurable, so this is a valid (if unwise) configuration that should work, not crash.

**Fix** (commit `a9a8848`): Skip payout when balance does not exceed the reserve.

**File**: `src/merchant/merchant.go`

---

## Bug 3: Stale timer race → permanent free internet — 🔴 FREE INTERNET

**What happens**: When a customer pays to extend their session, the old timer's callback can race with the new timer being stored. The stale callback deletes the **new** timer from the map. The customer stays authorized in ndsctl with no timer to ever close the gate — **permanent free internet**.

```
Timeline:
1. Customer pays → timerA created, stored in openGates[mac]
2. Customer pays again → timerB created, stored in openGates[mac], timerA.Stop() called
3. timerA's callback was already scheduled → fires → delete(openGates, mac)
4. timerB is now gone from the map. No timer will ever close the gate.
5. Customer has permanent free internet.
```

**How to trigger**: Customer pays twice in quick succession (extending their session). The race window is small but real under load or on slow hardware (like MIPS routers).

**Fix** (commit `a62f2f3`): Self-referential timer pointer (sentinel pattern). The callback captures its own pointer and only deletes the map entry if pointer equality holds. Stale callbacks skip the delete.

**Files**: `src/valve/valve.go`, `src/valve/valve_test.go`

---

## Tests (commit `6b37265`)

12 new regression tests: 8 merchant tests (PricePerStep=0, payout underflow), 4 valve tests (timer race).

## Files Changed

- `src/merchant/merchant.go` — guards for PricePerStep=0 and payout underflow
- `src/merchant/merchant_test.go` — 8 new tests
- `src/config_manager/config_schema.go` — Min: uint64(1) on PricePerStep
- `src/valve/valve.go` — sentinel timer pattern
- `src/valve/valve_test.go` — 4 new tests

## Merge Order

Independent of #163 (security hardening) and #164 (docs). Can merge in any order.

### DIFF
--- a/src/config_manager/config_schema.go
+++ b/src/config_manager/config_schema.go
@@ -56,7 +56,7 @@ func GetConfigSchema() []FieldSchema {
 				{Name: "BalanceTolerancePercent", JSONKey: "balance_tolerance_percent", Type: "uint64", Description: "Tolerance percentage for balance checks", Default: uint64(10), Required: true, Editable: true},
 				{Name: "PayoutIntervalSeconds", JSONKey: "payout_interval_seconds", Type: "uint64", Description: "Seconds between payout rounds", Default: uint64(60), Required: true, Editable: true},
 				{Name: "MinPayoutAmount", JSONKey: "min_payout_amount", Type: "uint64", Description: "Minimum payout amount in sats", Default: uint64(128), Required: true, Editable: true},
-				{Name: "PricePerStep", JSONKey: "price_per_step", Type: "uint64", Description: "Price per step in sats", Default: uint64(1), Required: true, Editable: true},
+				{Name: "PricePerStep", JSONKey: "price_per_step", Type: "uint64", Description: "Price per step in sats", Default: uint64(1), Required: true, Editable: true, Min: uint64(1)},
 				{Name: "PriceUnit", JSONKey: "price_unit", Type: "string", Description: "Price unit", Default: "sats", Required: true, Editable: true},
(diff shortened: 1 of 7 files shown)


Amperstrand:
Duplicate of #161 — same patches, different branch name. Closing this one.

PR #27, opened 2025-07-02
Fix: access point getting disconnected when installing tollgate package

c03rad0r:

PR #167, opened 2026-06-17
fix: multi-mint wallet registration + always open gate for bytes sessions

Amperstrand:
## Problem

Two bugs discovered during physical router testing (D-Link COVR-X1860, OpenWrt 24.10.2, nodogsplash 5.0.2).

### Bug 1: Only the first mint in config works

`tollwallet.New()` passes `acceptedMints[0]` as `CurrentMintURL` to `wallet.LoadWallet()`. The gonuts wallet supports multiple mints via `AddMint()`, but it was never called for the remaining mints. This causes `RequestMint()` (Lightning invoices) and `Receive()` (Cashu tokens) to fail with `"mint does not exist"` for any mint that isn't first in `accepted_mints`.

**Workaround used**: reorder `accepted_mints` in `config.json` to put the desired mint first. Fragile and confusing.

### Bug 2: Cashu payments return success but don't open the gate

`openGateForSession()` skips `OpenGate` when `valve.HasDataBaseline()` returns true, assuming the gate is already open. But after an `ndsctl deauth`, NDS closes the gate while the data baseline persists. Cashu payments return `kind=1022` (success) but the client stays `Preauthenticated` with no internet.

Lightning payments are unaffected because they're typically the first payment (no prior baseline exists).

### Multi-mint (tollwallet.go)
- **Startup**: calls `AddMint()` for all accepted mints during initialization
- **Lazy fallback**: `ensureMintRegistered()` retries `AddMint` on demand if a mint wasn't reachable at startup (handles mints that become available after WiFi connects)
- Tracks registered mints in a `map[string]bool` to avoid redundant network calls
- Thread-safe via `sync.Mutex`

### Gate-open (lightning.go)
- Removes the `HasDataBaseline` check — always calls `OpenGate`
- `ndsctl auth` is idempotent; re-authenticating an already-authenticated client is harmless

## Testing Evidence

Both bugs were discovered and fixed during end-to-end testing on physical hardware (D-Link COVR-X1860 A1 router, Motorola moto g8 plus phone).

**Without Fix 1** (multi-mint):
```
POST /ln-invoice {"amount":5,"mint_url":"https://nofee.testnut.cashu.space"}
→ {"error":"mint does not exist"}

# mint.coinos.io worked because it was acceptedMints[0]
POST /ln-invoice {"amount":5,"mint_url":"https://mint.coinos.io"}
→ {"quote":"...","invoice":"lnbc..."}  (success)
```

**With Fix 1** (any mint works regardless of order):
```
[...]

PR #165, opened 2026-06-17
feat(config): split dev profit-share across three maintainers

c03rad0r:
## Why

The default dev split paid both the `developer` (0.21) and `owner` (0.79) identities to the shared `tollgate@minibits.cash` address. This splits the dev cut so each of the three maintainers is paid to their own Lightning address.

## Changes

| Identity | Factor | Lightning address |
|---|---|---|
| `owner` (operator) | 0.79 | `tollgate@minibits.cash` (unchanged) |
| `c08r4d0r` | 0.07 | `c3e23eb5e3d00f18b2f4f588@coinos.io` |
| `amperstrand` | 0.07 | `npub1pu45ulqgtwetysp2ng8w5xyak8e3gp2vh2a4c952l88csuxatqpqmfny5l@lnurl.psbt.me` |
| `origami74` | 0.07 | `iwillnot@getalby.com` |

Factors sum to **1.00** (0.79 + 0.07×3), satisfying `ValidateProfitShare`.

## Files
- `src/config_manager/config_manager_identities.go` — replace `developer` identity with the three maintainers (each with its `lightning_address`)
- `src/config_manager/config_manager_config.go` — default `profit_share` now owner + three maintainers
- `README.md` + `docs/merchant.md` — `profit_share` JSON examples updated

`c08r4d0r` is the post-rename handle for c03rad0r (per PR #92).

## Verification
- `go build ./...` and `go test ./src/config_manager/...` pass
- No remaining `"developer"` identity references in the tree

PR #168, opened 2026-06-18
fix(merchant): owner-first robust profit-share payouts

c03rad0r:
## Why

Today, if **one** recipient's LNURL fails during a profit-share payout, `PayoutShare` calls `mintHealthTracker.MarkUnreachable(mint)` — faulting the entire funding mint (can trip degraded mode) even though only the *payee* failed. It also pays recipients in config order with no owner-first guarantee. With the 3-maintainer dev-split (#165), one dead maintainer LNURL could wrongly degrade the router.

This makes payouts robust: **owner paid first and must succeed**; recipients are probed for reachability; unreachable ones are skipped (funds stay); failures don't fault the mint.

## Algorithm
1. **Reachability probe** — fetch an invoice for every recipient (up to 5 retries each). Unreachable recipients are skipped; their share stays in the wallet (not redistributed, not tracked).
2. **Owner gate** — the owner (`identity == "owner"`) must be reachable AND paid successfully first. If the owner can't be reached or their payout fails → **abort the whole cycle** (no dev-split payouts); e-cash retained.
3. **Reachable maintainers** — paid independently; a maintainer's failure (after 5 melt retries) retains that maintainer's e-cash for the next cycle, doesn't affect the others, and does **not** fault the mint.

## Changes
- `processPayout`: rewritten (probe → owner gate → reachable maintainers).
- `PayoutShare`: returns an error; **removes `MarkUnreachable`** on payee-side failures — **also resolves #27** (PayoutShare no longer faults the mint on non-mint errors).
- `fetchInvoiceWithRetry` helper + `payoutInvoiceRetries = 5`.
- `MeltToLightning`: melt retry cap `10 → 5` (per the payout retry policy).

Mint-unreachability detection is unaffected — the health tracker's proactive probes still detect real mint outages independent of the payout path.

## Edge cases handled
- All maintainers unreachable, owner OK → owner paid, all dev shares retained.
- Owner reachable but every maintainer payout fails → owner paid, all dev e-cash retained, retried next cycle.

## Verification
- `go build ./...` and `go test ./src/merchant/... ./src/tollwallet/...` pass.
- (Pre-existing, unrelated) `go vet` warns at `merchant.go:147` about a `*TollWallet` copy introduced by #167's `sync.Mutex` — not from this PR; vet isn't a CI gate.
- Hardware E2E coverage: tracked in a companion test PR (linked shortly).

Follow-up (not in this PR): unit tests for `processPayout` would require a wallet/LNURL interface seam; per the repo's e2e-over-mocks direction, behavior is validated by the hardware tests.

### DIFF
--- a/src/merchant/merchant.go
+++ b/src/merchant/merchant.go
@@ -15,6 +15,7 @@ import (
 	"sync"
 
 	"github.com/OpenTollGate/tollgate-module-basic-go/src/config_manager"
+	"github.com/OpenTollGate/tollgate-module-basic-go/src/lightning"
 	"github.com/OpenTollGate/tollgate-module-basic-go/src/tollwallet"
 	"github.com/OpenTollGate/tollgate-module-basic-go/src/utils"
 	"github.com/OpenTollGate/tollgate-module-basic-go/src/valve"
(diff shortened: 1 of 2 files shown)


Amperstrand (approved):
## Approved — solid payment-correctness improvement

The owner-first payout model is correct:
1. **Reachability probe** before payout (fetch invoice, skip if unreachable) — prevents wasting ecash on failed melts
2. **Owner must be paid first** — if owner is unreachable, abort all payouts (ecash retained)
3. **Per-recipient error isolation** — a failed payout to one maintainer doesn't block others
4. **PayoutShare returns error** instead of marking mint unreachable — correctly separates payee failures from mint failures (resolves #27)

The melt retry reduction (10 → 5) is appropriate — 10 retries against a dead Lightning node just wastes time.

One observation for future consideration: the reachability probe fetches a real invoice for each recipient, which consumes a small amount of the recipient's Lightning quota. If payout cycles are frequent and recipients are many, consider caching the reachability result for a short TTL (e.g., 60s) to avoid probing on every cycle.

No blocking issues. Ship it.

Amperstrand (approved):
## Reviewed — looks good

**Core fix (correct):** `PayoutShare` no longer calls `mintHealthTracker.MarkUnreachable()` on payee failure. One recipient's LNURL being down should not fault the funding mint — that was a serious availability bug (#27).

**Three-phase payout (sound design):**
1. Reachability probe — fetch invoice with 5 retries per recipient
2. Owner-first — owner must be reachable and paid before any maintainer
3. Reachable maintainers — failed payouts retain e-cash for next cycle

**`PayoutShare` signature change** (void → error): callers can now distinguish payee failure from mint failure. Clean.

**Minor note (non-blocking):** The reachability probe fetches an invoice, then `MeltToLightning` fetches another during the melt. Double invoice fetch per recipient — latency cost but not a correctness issue.

Build: compiles clean. Ready to merge.

Issue #169, opened 2026-06-18
Tag-readiness: main @ 04ae54e → READY


c03rad0r:
Tag-readiness assessment of `main` @ `04ae54e77aebc2dab6bf9e6f58c6cb8dd6ce22a2` (post `v0.5.0-alpha2`).

## ⚠️ Verdict: READY-WITH-CAVEATS
- **OK for a `v0.5.0-alpha3` / `-beta1` tag.**
- **NOT for stable `v0.5.0`** until the caveats below are resolved.

Build is clean and healthy on real `mediatek/filogic` hardware; core API + reboot-recovery pass. The headline **v0.5.0 two-router features (upstream autopay / reseller) and degraded-mode-on-hardware were not exercised** because the second lab router is on an unsupported OpenWrt release.

| Tier | Result |
|------|--------|
| 0 — static (`go build`/`vet`/`test`) | ✅ build+vet clean; **12/14 modules pass**. Root-module test non-hermetic off-router (`/etc/tollgate/config.json`); `upstream_detector/go.mod` needs `go mod tidy`. |
| 1 — single-router API smoke (beta @ 04ae54e) | ✅ **27 passed / 24 skipped / 1 failed**. The 1 fail is env-only (`curl` not on the image); backend healthy. |
| 2 — two-router e2e (alpha+beta) | 🚫 **Not executed** — alpha (`192.168.8.1`) is stock OpenWrt **21.02** (GL-MT3000); can't run the package without a reflash. |
| 3 — reboot-recovery (beta) | ✅ **PASS** — clean cycle, service auto-started, build persisted. |

## To clear for stable v0.5.0
1. Reflash alpha to OpenWrt ≥24.10 and run `smoke-upstream` + reseller + `smoke-degraded` on hardware.
2. `go mod tidy` in `upstream_detector`; make the root-module test hermetic (temp config dir).
3. (test-automation repo) update `deploy.sh` (binary paths + dead LuCI files) and restore a fetchable artifact path (Blossom/Nostr or local `.ipk`).

## Artifacts
- Full report (committed): https://github.com/OpenTollGate/physical-router-test-automation/blob/feat/tag-readiness-suite/docs/tag-readiness-reports/TEST-REPORT-main-04ae54e.md
- Reproducible suite (Makefile + pytest): https://github.com/OpenTollGate/physical-router-test-automation/pull/38

[...]

PR #170, opened 2026-06-18
fix(valve): retry ndsctl auth to harden two-router autopay gate-open

c03rad0r:
## Problem

In the v0.5.0 reseller (router-to-router) flow, the upstream router's first gate-open after a payment fails:

```
Payment failed, attempting token recovery
error="payment rejected: failed to open gate: exit status 1"
❌ Renewal failed error="payment failed: payment rejected by upstream: failed to open gate: exit status 1"
```

It then recovers via the token-recovery path ~60–90s later. Verified end-to-end during the `main @ 04ae54e` tag-readiness campaign (issue #169) — the purchase works, but with this flaky first attempt.

## Root cause

The upstream's NoDogSplash does not always have the client session registered the instant the valve tries to authorize it. The session is created asynchronously by the captive-portal trigger in `upstream_session_manager/tollgate_prober.go` (the explicitly-labeled `TEMPORARY WORKAROUND`). That races with the valve's one-shot `ndsctl auth` in `authorizeMAC` → `OpenGate` returns the `exit status 1` → payment rejected.

## Fix

Make `authorizeMAC` retry `ndsctl auth` up to `authMaxAttempts` (5) with a short backoff. Authorize is idempotent, so a bounded retry is safe and removes the transient first-attempt failure.

- `runNdsctl` is now a stubbable package `var` so the retry is unit-testable without a real `ndsctl`.
- New tests: `TestAuthorizeMACRetriesThenSucceeds`, `TestAuthorizeMACFailsAfterMaxAttempts`. All existing valve tests still pass; `go build`/`go vet` clean.

## Scope / follow-up

This is the targeted reliability fix. The **proper long-term fix** (separate, larger PR) is to stop relying on the `tollgate_prober.go:106` `TEMPORARY WORKAROUND` portal-trigger entirely and have the upstream create the NDS client session inside the payment-acceptance path (idempotent). This PR makes the current architecture reliable; that follow-up retires the tech debt.

## Validation

Unit tests green. Hardware repro exists in the test-automation repo (`two_router_funded_upstream` fixture, PR [physical-router-test-automation#38](https://github.com/OpenTollGate/physical-router-test-automation/pull/38)) — once merged, that fixture will regression-guard this fix on real routers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,10 @@ Changes on `main` since `v0.4.0` (tagged 2026-04-06).
   division-by-zero, prevent a `uint64` underflow in payout, and stop a stale
   valve timer callback from deleting its replacement
   ([#161](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/161)).
+- **Two-router autopay reliability:** retry `ndsctl auth` briefly in the valve
+  so a payment's gate-open no longer fails on the first attempt when NoDogSplash
+  has not yet registered the reseller client (previously failed with "failed to
+  open gate" and recovered only via the token-recovery path ~60–90s later).
 - **BOLT11 / NoDogSplash:** make BOLT11 decode non-fatal and set the NoDogSplash
   gateway port to 2050
   ([#158](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/158)).
--- a/src/valve/valve.go
+++ b/src/valve/valve.go
@@ -17,9 +17,25 @@ import (
 // deadlocks (issue #387) that can cause ndsctl to hang indefinitely.
 const ndsctlTimeout = 5 * time.Second
 
+// authMaxAttempts bounds the number of authorizeMAC attempts.
+//
+// NoDogSplash does not always have a client session registered the instant we
+// try to authorize it — most notably in the two-router reseller flow, where the
+// upstream's NDS creates the client session asynchronously (see
+// upstream_session_manager/tollgate_prober.go's captive-portal trigger). Without
+// a retry the first payment attempt fails with "failed to open gate" and only
+// recovers via the token-recovery path ~60-90s later. The auth operation is
+// idempotent, so a bounded retry is safe.
+const authMaxAttempts = 5
+
+// authRetryDelay is the wait between authorizeMAC retries. It is a var so tests
+// can shrink it to keep the suite fast.
+var authRetryDelay = 400 * time.Millisecond
+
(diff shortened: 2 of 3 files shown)


c03rad0r:
## Hardware verification — the retry fires but does NOT fix the reseller gate-open

I deployed this branch (commit `978c1a0`) to the upstream router (beta) and re-ran the funded two-router purchase (`TestTwoRouterFunded` — alpha pays beta). The good news and the bad news:

**What works:** the retry logic is correctly built, deployed, and firing. beta valve logs show the new code path live:
```
valve: ndsctl auth failed, retrying (NoDogSplash may not have registered the client yet)  attempt=1 error="exit status 1" output="Failed to authenticate client 92:83:c4:8c:59:c4."
valve: ndsctl auth failed, retrying ... attempt=2 ...
valve: ndsctl auth failed, retrying ... attempt=3 ...
valve: ndsctl auth failed, retrying ... attempt=4 ...
valve: Error authorizing MAC address  error="exit status 1"
```
Unit tests pass and the funded purchase test still passes (no regression).

**What does NOT work:** the retry does **not** open the gate. NoDogSplash returns `"Failed to authenticate client <mac>"` on **every** attempt (all retries, across multiple payment cycles). `ndsctl auth` cannot authenticate a MAC that NDS does not yet have in its client list, so retrying it — whether 5×400ms (committed) or a 15s budget (I tried) — is ineffective. The gate eventually opens only because the **separate** `tollgate_prober.go:106` TEMPORARY-WORKAROUND captive-portal trigger eventually gets the reseller client into NDS, several payment cycles later — independent of this retry.

**Conclusion / recommendation:** this PR's `ndsctl auth` retry is a reasonable low-risk robustness improvement for cases where NDS is *almost* ready, but it does **not** fix the two-router gate-open failure, because the root cause is that the upstream's NDS does not register the reseller client before `OpenGate` is called. The actual fix is the one already noted in the PR description: **create the NDS client session inside the payment-acceptance path** (idempotent), or make the captive-portal trigger synchronously register the client before payment, and retire the `tollgate_prober.go:106` workaround. I'd suggest either (a) rework this PR toward that fix, or (b) merge it as a minor robustness tweak with this caveat clearly stated, and open a follow-up for the real fix. Happy to do either.

(Logs + the funded fixture are reproducible via [physical-router-test-automation#38](https://github.com/OpenTollGate/physical-router-test-automation/pull/38); tag-readiness context in [#169](https://github.com/OpenTollGate/tollgate-module-basic-go/issues/169).)

Issue #17, opened 2025-06-11
Broken config file when updating tollgate-basic


c03rad0r:
I end up in this state if I already had a config file and I update TollGate.

Temporary workaround: delete the config file and restart the `tollgate-basic` service.

```
Wed Jun 11 14:16:38 2025 daemon.err tollgate-basic[3004]: 2025/06/11 14:16:38 Failed to create config manager: json: cannot unmarshal object into Go struct field MintConfig.accepted_mints.url of type string
Wed Jun 11 14:16:43 2025 daemon.err tollgate-basic[3095]: 2025/06/11 14:16:43 Failed to create config manager: json: cannot unmarshal object into Go struct field MintConfig.accepted_mints.url of type string
Wed Jun 11 14:16:48 2025 daemon.err tollgate-basic[3257]: 2025/06/11 14:16:48 Failed to create config manager: json: cannot unmarshal object into Go struct field MintConfig.accepted_mints.url of type string
Wed Jun 11 14:16:53 2025 daemon.err tollgate-basic[3422]: 2025/06/11 14:16:53 Failed to create config manager: json: cannot unmarshal object into Go struct field MintConfig.accepted_mints.url of type string
Wed Jun 11 14:16:58 2025 daemon.err tollgate-basic[3616]: 2025/06/11 14:16:58 Failed to create config manager: json: cannot unmarshal object into Go struct field MintConfig.accepted_mints.url of type string
root@GL-E750:/tmp# cat /etc/tollgate/config.json | jq
{
  "config_version": "v0.0.2",
  "tollgate_private_key": "0cc7caf7db62ac262aee43cf693bd03aa277e3764f8b83bd7ed6591a26386aaf",
  "accepted_mints": [
    {
      "url": {
        "url": "https://mint.minibits.cash/Bitcoin",
        "min_balance": 2,
        "balance_tolerance_percent": 10,
        "payout_interval_seconds": 60,
        "min_payout_amount": 4
      },
      "min_balance": 100,
      "balance_tolerance_percent": 10,
      "payout_interval_seconds": 60,
      "min_payout_amount": 200
    },
    {
      "url": {
        "url": "https://mint2.nutmix.cash",
        "min_balance": 2,
        "balance_tolerance_percent": 10,
        "payout_interval_seconds": 60,
        "min_payout_amount": 4
      },
      "min_balance": 100,
      "balance_tolerance_percent": 10,
      "payout_interval_seconds": 60,
[...]

PR #180, opened 2026-06-29
fix(whoami): graceful degradation when MAC lookup fails

c03rad0r:
## Problem

When getMacAddress() fails (no DHCP lease, expired ARP entry, non-DHCP client, test environments), the /whoami handler returned HTTP 500 with an empty body. The captive-portal SPA parses the response as JSON; the empty 500 body yields undefined, causing TypeError: Cannot read properties of undefined (reading 'deviceInfo'). This crashes the entire portal SPA.

## Fix

Return HTTP 200 with JSON body when MAC lookup fails, mirroring the HandleBalance pattern. Also changed success path from plain-text to JSON for consistency.

## Test

TestWhoami_MACLookupFails_Returns200JSON verifies the fix.

Root cause identified during M0 UX demo recording.

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -339,13 +339,23 @@ func handler(w http.ResponseWriter, r *http.Request) {
 	var mac, err = getMacAddress(ip)
 
 	if err != nil {
-		mainLogger.WithError(err).Error("Error getting MAC address")
-		w.WriteHeader(http.StatusInternalServerError)
+		// Caller's IP is not in the DHCP lease file or ARP table — this
+		// happens when dnsmasq has just restarted, the lease expired, or the
+		// request comes from a non-DHCP/static-IP client. The captive-portal
+		// SPA parses the /whoami response as JSON; returning HTTP 500 with an
+		// empty body yields undefined on parse and crashes the SPA. Degrade
+		// gracefully instead: HTTP 200 with an empty MAC, mirroring the
+		// no-session behaviour already used by /balance (HandleBalance).
+		mainLogger.WithError(err).Debug("MAC lookup failed for /whoami, returning empty mac")
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		json.NewEncoder(w).Encode(map[string]string{"mac": ""})
 		return
 	}
 
 	mainLogger.WithField("mac", mac).Debug("MAC address resolved")
-	fmt.Fprint(w, "mac=", mac)
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(map[string]string{"mac": mac})
 }
 
 func handleDetails(w http.ResponseWriter, r *http.Request) {
--- a/src/whoami_test.go
+++ b/src/whoami_test.go
@@ -1,0 +1,66 @@
+package main
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestWhoami_MACLookupFails_Returns200JSON proves that GET /whoami degrades
+// gracefully when the caller's MAC address cannot be resolved (no DHCP lease
+// and no ARP entry). The captive-portal SPA parses the /whoami response as
+// JSON; an HTTP 500 with an empty body makes that parse yield undefined and
+// crashes the SPA. The endpoint must instead return HTTP 200 with a JSON body.
+//
+// Precondition: /tmp/dhcp.leases must be absent so getMacAddress() cannot
+// resolve the loopback caller. dnsmasq only writes that file on a router, so
+// it never exists on CI/dev machines; we guard explicitly to keep the test
+// deterministic and skip (rather than mis-report) when a real lease file is
+// present on a router-under-test.
+func TestWhoami_MACLookupFails_Returns200JSON(t *testing.T) {
+	if _, err := os.Stat("/tmp/dhcp.leases"); err == nil {
+		t.Skip("/tmp/dhcp.leases exists on this host; cannot deterministically " +
+			"trigger MAC-lookup failure — remove it to run this test")
+	}
+
+	// Sanity guard: confirm the failure path is actually reachable for the
+	// loopback address before asserting on the HTTP behaviour.
+	if _, err := getMacAddress("127.0.0.1"); err == nil {
+		t.Skip("getMacAddress resolved 127.0.0.1 on this host; cannot " +
+			"deterministically trigger the MAC-lookup failure path")
+	}
+
+	// Request originates from localhost. getIP() returns 127.0.0.1 (no
+	// forwarding headers), which getMacAddress() cannot resolve.
+	req := httptest.NewRequest(http.MethodGet, "/whoami", nil)
+	req.RemoteAddr = "127.0.0.1:1234"
+	w := httptest.NewRecorder()
+
+	// Invoke the same handler chain registered for the /whoami route in main().
+	CorsMiddleware(handler)(w, req)
+
+	// Must NOT be a 500 with an empty body (the bug).
+	assert.NotEqual(t, http.StatusInternalServerError, w.Code,
+		"/whoami must not return HTTP 500 when MAC lookup fails")
+	assert.NotEmpty(t, w.Body.String(),
+		"/whoami must not return an empty body when MAC lookup fails")
+
+	// Must be HTTP 200 with a JSON body the SPA can parse without crashing.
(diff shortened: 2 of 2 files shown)


c03rad0r:
## Related PRs

This is the **backend half** of the /whoami crash fix. The frontend defensive fix is also open:

- **tollgate-captive-portal-site #12** — Frontend null-guard on deviceInfo (prevents SPA crash regardless of backend state)
- **physical-router-test-automation #54** — Test infrastructure: auto-minting Cashu fixtures + log-based session verification (no token burning)

**No merge dependencies** — all three can be merged independently. But together they form the complete fix:
- This PR (#180): fixes the server-side root cause (500 empty body -> 200 JSON)
- PR #12: frontend defensive guard (SPA never crashes even if backend returns unexpected data)
- PR #54: regression test infrastructure (auto-minting fixtures, log-based verification)

c03rad0r:
## Motivation

While recording a demo video of the captive portal for our UX designer (M0), the entire portal showed a blank page instead of the payment UI. Root cause: the `/whoami` endpoint returns HTTP 500 with an empty body whenever the requesting client's MAC address can't be resolved from DHCP leases or the ARP table.

This happens in any scenario where the client IP isn't in dnsmasq state: after a dnsmasq restart, for static-IP clients, expired leases, or in test/CI environments without a real router. The React SPA tries to parse the empty 500 response as JSON, gets `undefined`, and crashes with `TypeError: Cannot read properties of undefined (reading 'deviceInfo')` — taking down the entire portal.

The fix is small (return JSON instead of crashing) and follows a pattern already established by `HandleBalance` in the same file.

Origami74:
The portal should not try to parse it. So the fix is not necessarily to change the response from the server.

I'm ok with always returning at least an empty json `{}`. and a different error code (404?). But i don't want it to respond with an ok status with empty mac.

Amperstrand (approved):
## Review + build verification

**Build**: ✅ `go build ./...` passes clean on main.

**The fix is correct**: when `getMacAddress()` fails (no DHCP lease, no ARP entry), return HTTP 200 with `{"mac":""}` instead of HTTP 500 with empty body. The captive-portal SPA parses `/whoami` as JSON — a 500 with empty body crashes the SPA's JSON parser.

**Response format change** (worth noting): `/whoami` now returns JSON `{"mac":"aa:bb:cc:dd:ee:ff"}` instead of plain text `mac=aa:bb:cc:dd:ee:ff`. This is a **breaking change** for any client parsing the old format. Verified that the captive-portal SPA ([tollgate-captive-portal-site](https://github.com/OpenTollGate/tollgate-captive-portal-site)) expects JSON from `/whoami` — so this actually fixes a format mismatch between the backend and the SPA.

**Test**: `whoami_test.go` covers the MAC-lookup-fails path. Test passes on OpenWrt (where `/tmp/dhcp.leases` is absent in CI). On macOS dev machines the test may fail because `getMacAddress` can resolve loopback via the ARP table — this is a platform-specific test issue, not a code bug.

**No issues found**. Ready to merge.

Amperstrand:
## Cross-repo concern: frontend `tollgate.js` still parses `/whoami` as plain text

While reviewing the related test infrastructure PR (physical-router-test-automation#54), I traced how the captive portal SPA consumes `/whoami` end-to-end. There appears to be a **breaking mismatch** between this PR and the frontend that PR #12 (captive-portal-site) does not address.

### Current frontend code (`src/helpers/tollgate.js` on main)

```javascript
const whoamiResponse = await fetch(`${baseUrl}/whoami`);

if (!whoamiResponse.ok) {
    return { status: "error", code: "TG002", ... };
}

// the response is expected to be in the format 'mac=00:11:22:33:44:55'
const whoamiText = await whoamiResponse.text();
const [identifierType, identifierValue] = whoamiText.trim().split("=");
const deviceInfo = { type: identifierType, value: identifierValue };
```

The frontend:
1. **Already checks `!whoamiResponse.ok`** — so a 4xx/5xx is handled (returns `TG002` error object). This is relevant to @Origami74's point below.
2. **Parses the body as plain text** with `.split("=")` — expects `mac=AA:BB:CC:DD:EE:FF`.

### What this PR (#180) changes

Both the success **and** error paths switch from plain text to JSON:

| Path | Before #180 | After #180 |
|------|------------|------------|
| Success | `mac=AA:BB:CC:DD:EE:FF` (text) | `{"mac":"AA:BB:CC:DD:EE:FF"}` (JSON) |
| Error | HTTP 500, empty body | HTTP 200, `{"mac":""}` (JSON) |

### The mismatch

After #180 merges, the success path breaks the frontend parser:

```
whoamiText = '{"mac":"AA:BB:CC:DD:EE:FF"}'
.split("=") → [{mac:AA:BB:CC:DD:EE:FF}]   // one element, no "=" found
identifierType  = {mac:AA:BB:CC:DD:EE:FF}
identifierValue = undefined
deviceInfo = { type: {mac:AA:BB:CC:DD:EE:FF}, value: undefined }
```

PR #12 (frontend null-guard) adds optional chaining in `DeviceInfo.jsx` and `cashu.js`, but **does not update `tollgate.js`** — the fetch/parse logic is untouched. So the `deviceInfo` object will have garbage in `.type` and `undefined` in `.value`, which the null-guards in #12 won't catch (the object exists, it's just wrong).

### Connection to @Origami74's concern

@Origami74 [commented](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/180#issuecomment-4833884873):
> "The portal should not try to parse it. I'm ok with always returning at least an empty json `{}`. and a different error code (404?). But I don't want it to respond with an ok status with empty mac."

The frontend **already handles non-OK responses** via `!whoamiResponse.ok`. So Origami74's 404 suggestion would work cleanly with the existing frontend code — the `TG002` error path fires, `DeviceInfo.jsx` renders without device info, no crash. The 200-with-empty-mac approach in this PR requires the frontend to check `mac === ""` as a special case, which is a weaker API contract.

### Options

1. **#180 returns 404 + `{}` on error, keeps text on success** — minimal change, frontend works as-is (error path fires, success path unchanged). Resolves Origami74's concern. No frontend changes needed at all.

2. **#180 returns JSON on both paths + update `tollgate.js` to `response.json()`** — requires a coordinated three-repo change (#180 backend JSON + #12 frontend null-guard + new `tollgate.js` parse update). Bigger surface area, more coordination.

3. **#180 as-is (200 + `{"mac":""}`) + update `tollgate.js`** — works but doesn't address Origami74's semantic concern about 200-for-not-found.

**My read:** Option 1 is the smallest, cleanest fix. The backend changes from `500 empty` → `404 {}`, the success path stays as text (`mac=...`), and the frontend's existing `!whoamiResponse.ok` check handles it. No frontend changes needed — which means #12 becomes optional hardening rather than a required dependency.

Happy to write up the exact diff for any of these options if that's helpful. Tagging @Origami74 since this directly connects to your comment.

Amperstrand:
## ⚠️ Escalation: success-path format change is a breaking change

Following up on my [earlier comment](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/180#issuecomment-4939440574) about the `tollgate.js` parsing mismatch — I want to emphasize that this is not just about the error path. **The success path also changes format**, and no frontend PR addresses it.

### The problem

This PR changes `/whoami` on **both** paths:

| Path | Before | After #180 |
|---|---|---|
| Error | HTTP 500, empty body | HTTP 200, `{"mac":""}` |
| **Success** | `mac=AA:BB:CC:DD:EE:FF` **(plain text)** | `{"mac":"AA:BB:CC:DD:EE:FF"}` **(JSON)** |

The frontend (`tollgate.js`) parses the success response as plain text:

```javascript
const whoamiText = await whoamiResponse.text();
const [identifierType, identifierValue] = whoamiText.trim().split("=");
```

After #180, `whoamiText` is `{"mac":"AA:BB:CC:DD:EE:FF"}`. Splitting on `=` produces one element — `identifierType` becomes the entire JSON string, `identifierValue` is `undefined`. The device info shown to the user is garbage.

### This needs a coordinated merge

If #180 merges without a corresponding frontend update:
- **Error path**: fixed (no more 500 empty body crash) — but see PR #17 for a separate frontend bug that also needs fixing
- **Success path**: **broken** — device info display shows malformed data

Options (unchanged from my earlier comment, restated for clarity):
1. **Don't change the success path format** — keep returning `mac=AA:BB:CC:DD:EE:FF` text on success, only fix the error path. No frontend changes needed.
2. **Change to JSON + update frontend** — requires a new frontend PR that replaces `.text().split("=")` with `.json()`. Coordinated merge across repos.

@Origami74 @c03rad0r — can we get a decision on the response format before this merges? The success-path breakage is currently unaddressed.

c03rad0r:
Closing — this work is already merged into the net4sats-mvp branch and will land on main when that branch is reviewed. No need to keep two copies open.

PR #20, opened 2025-06-14
Remove btclogv2 and resolve bugs/errors

c03rad0r:

PR #197, opened 2026-07-08
fix(merchant): credit token face value instead of post-fee amount

Amperstrand:
## Change

`merchant.go:439` now passes `paymentCashuToken.Amount()` (face value) to `calculateAllotment()` instead of `amountAfterSwap` (post-fee wallet balance).

## Why

The mint fee (`fee_ppk`) was silently passed to the customer — a 1-sat token on a `fee_ppk=10` keyset yielded **zero access**. The fee is an operator cost, not a customer cost.

## Alignment with Rust reference

tollgate-rs (maintained by @Origami74, 43/44 commits) already uses face value: `calculate_allotment(amount_sat, &state.config)`.

## Verification
- `go build -tags testenv ./...` — clean
- `go test -tags testenv -count=1 ./...` — all pass

Fixes Amperstrand/tollgate-module-basic-go#20.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,13 @@ and [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-Nothing yet.
+### Fixed
+
+- **Allotment based on token face value.** Payments now credit the token'"'"'s
+  face value (sum of proof amounts) rather than the post-fee wallet balance,
+  aligning with the Rust reference implementation. Users receive the full
+  amount they paid for; the mint fee is absorbed as an operator cost
+  ([#197](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/197)).
 
 ## [v0.5.0] - 2026-07-03
 
--- a/src/merchant/merchant.go
+++ b/src/merchant/merchant.go
@@ -432,11 +432,10 @@ func (m *Merchant) PurchaseSession(cashuToken string, macAddress string) (*nostr
 		return noticeEvent, nil
 	}
 
-	log.Printf("Amount after swap: %d", amountAfterSwap)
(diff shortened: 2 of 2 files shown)


Amperstrand:
## Code review (can't self-approve — posting as comment)

The fix is correct and minimal: use `paymentCashuToken.Amount()` (face value) instead of `amountAfterSwap` (post-fee). Customer gets the full allotment they paid for; mint fee absorbed as operator cost. Matches the Rust reference. Builds clean. Resolves Amperstrand #20.

Amperstrand:
Closing this PR — the change is incorrect for the Go implementation.

## Why

The Rust reference impl credits **face value** because it uses `BootstrapWallet::verify()` which only calls NUT-07 checkstate (a read-only query). It does NOT swap/receive the token, so no fee is paid. Face value is correct in that context.

The Go impl uses `TollWallet.Receive()` which calls the mint's `/v1/swap` endpoint. The mint charges `fee_ppk` during the swap. The returned `amountAfterSwap` is the post-fee balance. Crediting face value while paying the fee would mean the operator subsidizes every transaction.

## The correct Cashu merchant pattern

The Cashu merchant docs say:
> "If the received Cashu token is valid, always swap it before treating the transaction as complete."
> "Remember to account for swap and melt fees."

The Go impl follows this guidance. The Rust `BootstrapWallet` does not (it only verifies).

## The 1-sat edge case

The original issue (#20) reported that 1-sat tokens yield zero access when fee_ppk > 0. This is a **pricing problem**, not an allotment-calculation problem. Solutions:

1. Set `min_purchase_steps` per-mint so post-fee amount always yields ≥1 step
2. Raise `price_per_step` to 2 (the workaround from the issue body)
3. Add a minimum-payment rejection for tokens where post-fee < price_per_step

These are UX decisions that belong in a separate PR.

PR #199, opened 2026-07-09
feat(wgm): re-implement vendor IE discovery from wire format spec

Amperstrand:
## What changed

Replaces the old dead-code bypass in `vendor_element_manager.go` (which referenced undefined `bitcoinOUI`/`nostrOUI` from a previous iteration) with a clean TollGate vendor IE implementation.

## Implementation

Wire format: `DD <len> 212121 01 <version> <flags> [TLV mint_url] [TLV pubkey]`

| Function | Purpose |
|---|---|
| `EncodeTollGateVendorIE` | Builds vendor-specific IE hex string from TollGateAdvertisement |
| `ParseTollGateVendorIE` | Extracts TollGateAdvertisement from raw IE bytes |
| `ParseVendorIEsFromScanData` | Walks raw 802.11 IE bytes from `iw` scan output |
| `EmitTollGateVendorIE` | Emits via `ubus set_vendor_elements` on hostapd interfaces |

**Config:** `vendor_ie_discovery` flag in `upstream_wifi` section (default `false`).
**Types:** `TollGateAdvertisement` struct, `IsTollGate` field on `NetworkInfo`, `VendorIEDiscovery` on `UpstreamManagerConfig`.
**Connector:** `ExecuteUbus` method added.

## Context

The feature was originally implemented in commit `70e4320` (on `feature/vendor-ie-discovery` branch) but never merged. That branch fell 24 commits behind v0.5.0. This PR re-implements from scratch on current main, using the wire format from the original commit message as the spec.

## What is NOT included (follow-up work)

- Scanner integration: populating `RawIEs` from `iw phy` scan data
- `main.go` integration: calling `EmitTollGateVendorIE` on startup
- `upstream_manager` integration: using `IsTollGate` in `findResellerCandidates`

These wiring changes can land in a follow-up PR once the core encode/decode is reviewed.

## Verification
- `go build` in wireless_gateway_manager module — clean
- `go build -tags testenv` in root module — clean

Fixes Amperstrand/tollgate-module-basic-go#47.

### DIFF
--- a/src/wireless_gateway_manager/vendor_element_manager.go
+++ b/src/wireless_gateway_manager/vendor_element_manager.go
@@ -1,123 +1,230 @@
-// Package wireless_gateway_manager implements the VendorElementProcessor for handling Bitcoin/Nostr vendor elements.
 package wireless_gateway_manager
 
 import (
+	"encoding/hex"
+	"fmt"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 )
 
 const (
-	tollgateOUI      = "212121" // TollGate custom OUI
-	tollgateElemType = "01"     // TollGate custom elementType
+	tollgateOUI          = "212121"
+	tollgateElemType     = "01"
+	tollgateElemTypeByte byte = 0x01
+
+	flagIsReseller  = 0x01
+	flagHasInternet = 0x02
+	flagOpenNetwork = 0x04
+
+	tlvTypeMintURL = 0x01
+	tlvTypePubkey  = 0x02
 )
 
-// ExtractAndScore extracts vendor elements from NetworkInfo and calculates a score.
-func (v *VendorElementProcessor) ExtractAndScore(ni NetworkInfo) (map[string]interface{}, int, error) {
-	// Temporarily bypass vendor element parsing as per user request
-	// rawIEs := ni.RawIEs
-	vendorElements := make(map[string]interface{}) // Initialize as empty for now
-	/*
-		vendorElements, err := v.parseVendorElements(rawIEs)
-		if err != nil {
-			v.log.Printf("[wireless_gateway_manager] ERROR: Failed to parse vendor elements: %v", err)
-			return nil, 0, err
-		}
-	*/
+func EncodeTollGateVendorIE(adv TollGateAdvertisement) (string, error) {
+	var flags uint8
+	if adv.IsReseller {
+		flags |= flagIsReseller
+	}
+	if adv.HasInternet {
+		flags |= flagHasInternet
+	}
+	if adv.OpenNetwork {
+		flags |= flagOpenNetwork
+	}
 
+	oui, err := hex.DecodeString(tollgateOUI)
+	if err != nil {
+		return "", fmt.Errorf("invalid OUI hex: %w", err)
+	}
+
+	elemType, _ := hex.DecodeString(tollgateElemType)
+	body := append(oui, elemType...)
+	body = append(body, adv.Version, flags)
+
+	if adv.MintURL != "" {
+		mintBytes := []byte(adv.MintURL)
+		if len(mintBytes) > 255 {
+			return "", fmt.Errorf("mint_url too long: %d bytes (max 255)", len(mintBytes))
+		}
+		body = append(body, tlvTypeMintURL, uint8(len(mintBytes)))
+		body = append(body, mintBytes...)
+	}
+
+	if len(adv.Pubkey) > 0 {
+		if len(adv.Pubkey) > 255 {
+			return "", fmt.Errorf("pubkey too long: %d bytes (max 255)", len(adv.Pubkey))
+		}
+		body = append(body, tlvTypePubkey, uint8(len(adv.Pubkey)))
+		body = append(body, adv.Pubkey...)
+	}
+
+	ie := append([]byte{0xDD, uint8(len(body))}, body...)
+
+	return hex.EncodeToString(ie), nil
+}
+
+func ParseTollGateVendorIE(raw []byte) *TollGateAdvertisement {
+	if len(raw) < 8 {
+		return nil
+	}
+	if raw[0] != 0xDD {
+		return nil
+	}
+
+	bodyLen := int(raw[1])
+	if len(raw) < 2+bodyLen {
+		return nil
+	}
+	body := raw[2:]
+
+	oui := fmt.Sprintf("%02x%02x%02x", body[0], body[1], body[2])
+	if oui != tollgateOUI {
+		return nil
+	}
+	if body[3] != tollgateElemTypeByte {
+		return nil
+	}
+
+	adv := &TollGateAdvertisement{
+		Version: body[4],
+	}
+
+	if len(body) > 5 {
+		flags := body[5]
+		adv.IsReseller = flags&flagIsReseller != 0
+		adv.HasInternet = flags&flagHasInternet != 0
+		adv.OpenNetwork = flags&flagOpenNetwork != 0
+	}
+
+	tlvOffset := 6
+	for tlvOffset+2 <= len(body) {
+		tlvType := body[tlvOffset]
+		tlvLen := int(body[tlvOffset+1])
+		if tlvOffset+2+tlvLen > len(body) {
+			break
+		}
+		tlvValue := body[tlvOffset+2 : tlvOffset+2+tlvLen]
+
+		switch tlvType {
+		case tlvTypeMintURL:
+			adv.MintURL = string(tlvValue)
+		case tlvTypePubkey:
+			adv.Pubkey = make([]byte, len(tlvValue))
+			copy(adv.Pubkey, tlvValue)
+		}
+
+		tlvOffset += 2 + tlvLen
+	}
+
+	return adv
+}
+
+func ParseVendorIEsFromScanData(raw []byte) []*TollGateAdvertisement {
+	var results []*TollGateAdvertisement
+
+	offset := 0
+	for offset+1 < len(raw) {
+		elemID := raw[offset]
+		length := int(raw[offset+1])
+
+		if offset+2+length > len(raw) {
+			break
+		}
+
+		if elemID == 0xDD && length >= 4 {
+			ieBytes := raw[offset : offset+2+length]
+			if adv := ParseTollGateVendorIE(ieBytes); adv != nil {
+				results = append(results, adv)
+			}
+		}
+
+		offset += 2 + length
+	}
+
+	return results
+}
+
+func (v *VendorElementProcessor) ExtractAndScore(ni NetworkInfo) (map[string]interface{}, int, error) {
+	vendorElements := make(map[string]interface{})
 	score := v.calculateScore(ni, vendorElements)
 	return vendorElements, score, nil
 }
 
-/*
-func (v *VendorElementProcessor) parseVendorElements(rawIEs []byte) (map[string]interface{}, error) {
-	vendorElements := make(map[string]interface{})
-
-	// Ensure rawIEs has at least 3 bytes for OUI
-	if len(rawIEs) < 3 {
-		return nil, fmt.Errorf("rawIEs too short to parse OUI: %d bytes", len(rawIEs))
-	}
-
-	oui := hex.EncodeToString(rawIEs[:3])
-	if strings.Contains(bitcoinOUI, oui) || strings.Contains(nostrOUI, oui) {
-		data := rawIEs[3:]
-		// Ensure data has at least 8 bytes for kbAllocation and contribution
-		if len(data) < 8 {
-			return nil, fmt.Errorf("vendor element data too short: %d bytes, expected at least 8", len(data))
-		}
-		kbAllocation, err := strconv.ParseFloat(string(data[:4]), 64)
-		if err != nil {
-			return nil, err
-		}
-		contribution, err := strconv.ParseFloat(string(data[4:8]), 64)
-		if err != nil {
-			return nil, err
-		}
-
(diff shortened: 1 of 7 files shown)


Amperstrand:
## Reviewed — well-structured feature, gated behind config flag

**Feature:** Re-implements TollGate vendor IE (802.11 vendor-specific IE) encoding/decoding from the wire format spec. Adds `EncodeTollGateVendorIE`, `ParseTollGateVendorIE`, `ParseVendorIEsFromScanData`, and `EmitTollGateVendorIE` (ubus emission).

**Good design choices:**
- Gated behind `vendor_ie_discovery` config flag (default `false`) — opt-in, no impact on existing deployments
- `ExecuteUbus` added to Connector — clean abstraction
- `TollGateAdvertisement` struct with clear fields (Version, IsReseller, HasInternet, OpenNetwork, MintURL, Pubkey)
- `IsTollGate` field on `NetworkInfo` — easy filtering in scan results
- Removes old dead-code bypass

**Vendor IE format:** `DD <len> 212121 01 <version> <flags> [TLV mint_url] [TLV pubkey]` — matches the OUI `21:21:21` (TollGate). TLV encoding for variable-length fields.

**Not fully reviewed:** I only saw the first ~120 lines of the diff. The full 441 lines include the encoding/decoding implementation which should be verified against the wire format spec. Recommend @c0brad0r or @Origami74 review the hex parsing and bounds checking in `ParseTollGateVendorIE`.

No blocking issues found in the reviewed portion.

c03rad0r (commented):
Thanks for implementing this — the vendor IE encoding/decoding is well-structured and the test coverage is thorough (18 cases covering roundtrips, flag combinations, edge cases, multi-IE parsing, scoring).

Not approving yet pending discussion on SSID/vendor-IE strategy (see separate comment below).

Three concerns worth addressing before merge:

**1. OUI 0x212121 (unregistered)**
Using an unregistered OUI means real Wi-Fi chipsets may not pass it through to userspace scan results. The PR body correctly notes this needs hardware testing. Has this been validated on any target hardware (ath79, mt7622, ipq806x)? If chipset drivers filter unknown OUIs, the entire feature is dead on arrival.

**2. Silent ubus error swallowing**
SetLocalAPVendorElements logs a warning and returns nil on ubus failure. If hostapd doesn't support vendor_elements via ubus (some OpenWrt builds don't compile it in), IE emission fails silently. Acceptable behind a default-off flag, but consider adding a startup health check that verifies ubus vendor_elements is supported when the flag is enabled.

**3. ExecuteUbus JSON via fmt.Sprintf**
The JSON payload in SetLocalAPVendorElements is built via fmt.Sprintf rather than encoding/json. For the current single key-value structure this is safe, but it will break if vendor_elements values ever contain special characters. Consider using encoding/json for robustness.

Happy to see this land once we resolve the SSID question. Great work.

c03rad0r:
## SSID strategy question — should we change SSID format alongside vendor IE?

Now that vendor IE can carry gateway metadata (version, flags, mint URL, pubkey), is this a good time to simplify the SSID? Current format: `TollGate-ABCD-2.4GHz` / `TollGate-ABCD-5GHz`

### Options

**Option A: Remove random chars → `TollGate-2.4GHz` / `TollGate-5GHz`**
- All TollGates share the same SSID → clients roam between them automatically
- DANGEROUS: each TollGate is independently operated with different pricing and mints. WiFi roaming assumes all APs lead to the same network. A client roaming from Gateway 1 ($0.01/MB) to Gateway 2 ($0.05/MB) mid-session loses their paid session and must re-pay at the new rate.
- **Not recommended.**

**Option B: Keep random chars, remove band suffix → `TollGate-ABCD`** ⭐
- Each gateway still uniquely identifiable by SSID
- 2.4GHz and 5GHz radios for the same gateway share one SSID → device auto-selects best band (band steering)
- Pro: cleaner SSID, removes redundant band info that the device already knows from the beacon
- Pro: no roaming risk — each gateway has a distinct SSID
- Pro: vendor IE carries the metadata that the band suffix used to encode
- **Recommended as next step.**

**Option C: Remove everything → `TollGate`**
- All TollGates share SSID, distinguishable only via BSSID + vendor IE
- Same roaming problem as Option A
- Would require client-side changes to prevent unwanted hops between gateways
- **Not recommended without client-side hop prevention.**

### My recommendation

This PR should stay focused on vendor IE encoding/decoding. SSID format change is a separate concern — it affects the setup script (`99-tollgate-setup`), the captive portal, and potentially client roaming behavior. It should be a separate PR.

If we want to act on the SSID simplification, **Option B** (remove band suffix, keep random chars) is the safest first step. We can revisit Option C once we have client-side logic that reads vendor IEs and prevents roaming to a gateway with incompatible pricing/mint.

src/wireless_gateway_manager/vendor_element_manager.go
```diff
@@ -1,123 +1,230 @@
...
-	return vendorElements, score, nil
+	if len(adv.Pubkey) > 0 {
+		if len(adv.Pubkey) > 255 {
+			return "", fmt.Errorf("pubkey too long: %d bytes (max 255)", len(adv.Pubkey))
+		}
+		body = append(body, tlvTypePubkey, uint8(len(adv.Pubkey)))
+		body = append(body, adv.Pubkey...)
+	}
+
+	ie := append([]byte{0xDD, uint8(len(body))}, body...)
```
c03rad0r, line 62: 🔴 **Encoder total-body-length overflow.** `uint8(len(body))` silently truncates if the combined body exceeds 255 bytes. The per-field checks above (mint ≤ 255, pubkey ≤ 255) don't prevent the *combined* overflow, and 802.11 caps the IE body at 255.

Reproducer (passes against current code, proving the truncation):
```go
adv := TollGateAdvertisement{
    Version: 1,
    MintURL: string(make([]byte, 200)),
    Pubkey:  make([]byte, 100),
}
ieHex, _ := EncodeTollGateVendorIE(adv)
raw, _   := hex.DecodeString(ieHex)
// raw[1] == 54  (310 % 256), not 310 — silent truncation
```

Suggested fix:
```go
if len(body) > 255 {
    return "", fmt.Errorf("vendor IE body too long: %d bytes (max 255)", len(body))
}
ie := append([]byte{0xDD, uint8(len(body))}, body...)
```
A test exercising the combined overflow (mint+pubkey > 255) would close the gap left by the two per-field `*TooLong` tests.

c03rad0r (commented):
Careful pass on the parser/encoder logic — this is the follow-up to the earlier partial review that deferred the hex-parsing and bounds-checking verification. Two correctness bugs in the wire-format length handling, both verified with executable test cases, plus a stale module doc. The overall structure (clean separation of encode / parse / scan-walk / emit, flag-gated, TLV-forward-compatible) is sound; the issues are narrowly in length handling.

## Encoder: total body length can overflow the 1-byte length field

`EncodeTollGateVendorIE` validates each field individually (mint ≤ 255, pubkey ≤ 255) but never checks the **combined** body length before writing `uint8(len(body))`. An 802.11 vendor-specific IE body is capped at 255 bytes (the Length field is a single octet). A 200-byte mint URL + 100-byte pubkey produces a 310-byte body; `uint8(310)` silently truncates to **54**, emitting a malformed IE that declares 54 bytes but carries 310. Decoding that IE yields garbage mint/pubkey — and since this is the wire format other TollGate routers must match, a corrupted frame on one node poisons discovery on its neighbors. See inline + the suggested guard + a test covering the combined-overlap branch (the existing `TestEncode_MintURLTooLong` / `TestEncode_PubkeyTooLong` only cover per-field limits).

## Parser: declared body length is ignored after the bounds check

`ParseTollGateVendorIE` validates `len(raw) >= 2+bodyLen` but then slices `body := raw[2:]`, discarding `bodyLen` and absorbing any trailing bytes into the parse. The internal `ParseVendorIEsFromScanData` caller slices each IE tightly so it doesn't trigger this today, but `ParseTollGateVendorIE` is exported and a direct caller with a non-tight buffer gets wrong results. Defense-in-depth fix inline.

## Stale module doc

`docs/wireless_gateway_manager.md` still says "Extracts and scores vendor-specific IEs (Bitcoin/Lightning info)" — the Bitcoin/Lightning OUI code was removed by this PR. The config table in the same file is also missing the new `vendor_ie_discovery` row. CHANGELOG entry is good; the module doc that describes this exact component is the gap.

## Minor

`SetLocalAPVendorElements` builds `combinedHex` by ranging over a `map[string]string`, so multi-element ordering is non-deterministic. Single caller today (`EmitTollGateVendorIE` passes one entry), so dormant — worth a note for the future multi-IE caller.

## Looks good

- TLV loop bounds (`tlvOffset+2 <= len(body)` + value-range check before slicing) are correct — no OOB reads.
- Pubkey copied via `make`+`copy`, avoiding slice aliasing into the input buffer.
- Unknown TLV types silently skipped — forward-compatible with future fields.
- `ExecuteUbus` mirrors the existing `ExecuteUCI` pattern; graceful no-op when hostapd/ubus absent is right for a flag-gated feature.
- Test coverage is genuinely good: hand-crafted byte sequences for the rejection paths (wrong OUI, wrong type, truncated TLV), not just round-trips.

## Disposition

Recommend **request changes** for the two length-handling bugs (encoder overflow + parser trailing-bytes) and the doc update. All three are small, mechanical fixes; once addressed this is a clean land.

src/wireless_gateway_manager/vendor_element_manager.go
```diff
@@ -1,123 +1,230 @@
...
+	}
 
-	// Ensure rawIEs has at least 3 bytes for OUI
-	if len(rawIEs) < 3 {
-		return nil, fmt.Errorf("rawIEs too short to parse OUI: %d bytes", len(rawIEs))
+	bodyLen := int(raw[1])
+	if len(raw) < 2+bodyLen {
+		return nil
 	}
+	body := raw[2:]
```
c03rad0r, line 79: ⚠️ **Parser ignores declared `bodyLen` after the bounds check.** `body := raw[2:]` absorbs any trailing bytes into the parse instead of honoring the declared length. The internal `ParseVendorIEsFromScanData` caller slices each IE tightly so this is dormant today, but `ParseTollGateVendorIE` is exported — a direct caller passing a padded buffer or full beacon blob gets wrong results.

Reproducer (declared bodyLen=5, one trailing byte):
```go
raw := []byte{
    0xDD, 0x05,               // declared body length 5 (OUI+type+version, NO flags)
    0x21, 0x21, 0x21, 0x01, 0x01,
    0x02,                      // trailing byte — outside declared body
}
parsed := ParseTollGateVendorIE(raw)
// parsed.HasInternet == true  — trailing 0x02 read as the flags byte
```

Suggested fix (defense-in-depth):
```go
body := raw[2 : 2+bodyLen]
```

c03rad0r:
Consolidated change-request list (merging the two reviews above so nothing gets lost):

**Correctness (parser/encoder):**
1. Encoder total-body-length overflow — `uint8(len(body))` silently truncates when mint+pubkey exceed 255 bytes combined (per-field checks do not guard the sum). Reproducer + fix in the review above.
2. Parser ignores declared `bodyLen` — `body := raw[2:]` reads trailing bytes as flags/TLVs. Fix: `body := raw[2 : 2+bodyLen]`.

**Strategy / robustness:**
3. OUI `0x212121` is unregistered — needs hardware validation that chipsets pass it through to userspace scan results (ath79 / mt7622 / ipq806x).
4. `SetLocalAPVendorElements` swallows ubus errors — consider a startup health check for `vendor_elements` ubus support when the flag is on.
5. JSON payload built via `fmt.Sprintf` — safe today (hex-only input), switch to `encoding/json` if values ever broaden.

**Docs:**
6. `docs/wireless_gateway_manager.md` still describes "Bitcoin/Lightning info" (removed by this PR) and the config table is missing `vendor_ie_discovery`.

**Minor:**
7. `combinedHex` map-range ordering is non-deterministic for multi-IE callers (single caller today).

Items 1, 2, and 6 are small mechanical fixes; 3 is the open hardware-validation question. Once those are addressed this is a clean land.

c03rad0r:
Formalized the SSID format discussion above as **#229** for tracking and maintainer consensus. This PR stays focused on vendor IE encode/decode — the SSID change (if approved) will be a separate PR.

Amperstrand:
## Change requests addressed (items 1, 2, 6)

Pushed fixes for the three mechanical items:

**1. Encoder body-length overflow** — Added sum-length guard after all TLVs are appended:
```go
if len(body) > 255 {
    return "", fmt.Errorf("vendor IE body too long: %d bytes (max 255)", len(body))
}
```

**2. Parser ignores bodyLen** — Fixed:
```go
// Before: body := raw[2:]
body := raw[2 : 2+bodyLen]
```

**3. Docs reference** — Updated `docs/wireless_gateway_manager.md`:
```
- VendorElementProcessor — Encodes/decodes TollGate vendor-specific IEs (802.11 IE type 221, OUI 21:21:21).
  Carries gateway metadata: version, flags, mint URL, pubkey. Gated behind vendor_ie_discovery config flag.
```
(Removed stale "Bitcoin/Lightning info" reference.)

**Items not addressed in this PR (per your guidance):**
- **3 (OUI hardware validation)** — open question, needs ath79/mt7622/ipq806x testing
- **4 (ubus health check)** — enhancement, can be follow-up
- **5 (JSON via fmt.Sprintf)** — safe today (hex-only), switch if values broaden
- **7 (map ordering)** — single caller today, non-issue

Build + all wireless_gateway_manager tests pass. @c03rad0r

Amperstrand:
Closing fork-based PR to recreate as same-repo PR. Change requests from c03rad0r (items 1, 2, 6) are addressed in the new PR.

PR #200, opened 2026-07-09
feat(valve): post-payment redirect with configurable auth delay

Amperstrand:
## What changed

Rebased the pre-v0.5.0 `feat/post-payment-redirect` branch (7 commits) onto current main. Conflicts in `valve.go` resolved by merging upstream'"s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,14 @@ and [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
-Nothing yet.
+### Added
+
+- **Post-payment redirect with configurable auth delay.** Rebased from
+  the pre-v0.5.0 feature branch onto current main. Adds
+  `auth_delay_seconds` and `redirect_url` config fields, delayed-auth
+  goroutine lifecycle in the valve (cancellable on CloseGate), and a
+  `welcome.html` captive-portal page shown after payment
+  ([#200](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/200)).
 
 ## [v0.5.0] - 2026-07-03
 
--- a/packaging/files/tollgate-captive-portal-site/balance.html
+++ b/packaging/files/tollgate-captive-portal-site/balance.html
@@ -20,5 +20,11 @@
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
     <div id="root"></div>
+    <script>
+      setTimeout(function() {
+        var params = window.location.search;
+        window.location.replace('/welcome.html' + params);
+      }, 3000);
+    </script>
   </body>
 </html>
--- a/packaging/files/tollgate-captive-portal-site/welcome.html
+++ b/packaging/files/tollgate-captive-portal-site/welcome.html
@@ -1,0 +1,131 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" />
+    <title>TollGate — Welcome</title>
+    <style>
+        * { margin: 0; padding: 0; box-sizing: border-box; }
+        body {
+            font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, sans-serif;
+            background: #0a0a0a;
+            color: #fff;
+            min-height: 100vh;
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            justify-content: center;
+            padding: 2rem 1rem;
+        }
+        h1 { font-size: 1.5rem; margin-bottom: 0.5rem; text-align: center; }
+        .subtitle { color: #999; margin-bottom: 2rem; text-align: center; }
+        .countdown {
+            font-size: 4rem;
+            font-weight: 200;
+            color: #4CAF50;
+            margin-bottom: 1rem;
+        }
+        .status { font-size: 0.85rem; color: #666; }
+        .btn {
+            display: inline-block;
+            padding: 0.75rem 1.5rem;
+            border-radius: 8px;
+            font-size: 1rem;
+            font-weight: 600;
+            text-decoration: none;
+            text-align: center;
+            cursor: pointer;
+            border: none;
+            background: #4CAF50;
+            color: #fff;
+            margin-top: 1rem;
+        }
+        #log {
+            display: none;
+            width: 100%;
+            max-width: 400px;
+            margin-top: 1rem;
+            padding: 0.75rem;
+            background: #111;
+            border-radius: 8px;
+            font-family: monospace;
+            font-size: 0.7rem;
+            color: #666;
+            max-height: 120px;
+            overflow-y: auto;
+            white-space: pre-wrap;
+            word-break: break-all;
+        }
+        #log.visible { display: block; }
+    </style>
+</head>
+<body>
+    <h1>You're connected!</h1>
+    <p class="subtitle">Redirecting you shortly...</p>
+    <div class="countdown" id="countdown">5</div>
+    <button class="btn" id="btn-now">Go Now</button>
+    <div class="status" id="status">Counting down...</div>
+    <div id="log"></div>
+
+    <script>
+        (function() {
+            var DEFAULT_TARGET = 'https://wallet.cashu.me/welcome';
+            var params = new URLSearchParams(window.location.search);
+            var TARGET = params.get('redirect') || DEFAULT_TARGET;
+            var DEBUG = params.has('debug');
+
+            var log = document.getElementById('log');
+            var countdownEl = document.getElementById('countdown');
+            var statusEl = document.getElementById('status');
+            var btnNow = document.getElementById('btn-now');
+
+            if (DEBUG) log.classList.add('visible');
+
+            function addLog(msg) {
+                log.textContent += new Date().toISOString().substr(11, 8) + ' ' + msg + '\n';
+                log.scrollTop = log.scrollHeight;
+            }
+
+            function toIntentUrl(url) {
+                var u = new URL(url);
+                var host = u.host;
+                var path = u.pathname + u.search + u.hash;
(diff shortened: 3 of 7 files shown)


Amperstrand:
## Reviewed — feature looks good, one security concern

**Feature:** Post-payment redirect with configurable auth delay. Adds `auth_delay_seconds` and `redirect_url` config fields, delayed-auth goroutine (cancellable on CloseGate), and a `welcome.html` page with countdown timer.

**Architecture (sound):** The delayed-auth goroutine in the valve is properly cancellable via `CloseGate`. The `welcome.html` page is standalone static HTML with a countdown and "Go Now" button.

**⚠️ Open redirect vulnerability in `welcome.html`:**
```javascript
var TARGET = params.get('redirect') || DEFAULT_TARGET;
window.location.replace(TARGET);
```
The `redirect` URL parameter is user-controlled and passed directly to `window.location.replace()`. An attacker could craft `http://router:2050/welcome.html?redirect=https://evil.com` to redirect users after payment.

**Fix:** validate the redirect URL against an allowlist, or only allow relative paths:
```javascript
var TARGET = params.get('redirect');
if (TARGET && !TARGET.startsWith('/')) TARGET = DEFAULT_TARGET;
```

**Non-blocking but worth noting:** `balance.html` now auto-redirects to `welcome.html` after 3 seconds via `setTimeout`. This changes the balance page behavior for all users, not just post-payment. Is that intentional?

Needs @c0brad0r or @Origami74 to review.

c03rad0r:
## Review — #200 post-payment redirect with configurable auth delay

This PR adds a two-part UX feature: after a client pays, the captive
portal shows a `welcome.html` countdown page, and the valve defers the
`ndsctl auth` call by a configurable delay so the portal page has time
to load before Android detects connectivity and closes the WebView.
New config fields `redirect_url` (string) and `auth_delay_seconds`
(int, 0–300) gate the behavior. CI is green across all matrix jobs,
`gofmt`/`vet`/`build-purity`/`contract-lint` all pass, and I re-ran
the valve and main-package test suites locally with `-race` — clean.

### The valve logic is sound

The core concurrency design is the strongest part of this PR. The new
`pendingUntil` map
([valve.go#L65](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/valve/valve.go#L65))
correctly handles the "concurrent payment during the delay window" race:
if a second `OpenGateUntil` arrives while a delayed auth is pending, it
updates `pendingUntil[mac]` in place and returns, and the goroutine
reads the latest value when it wakes
([valve.go#L411-L415](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/valve/valve.go#L411-L415)).
`CloseGate` deletes the pending entry
([valve.go#L311](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/valve/valve.go#L311)),
so the goroutine aborts cleanly via the `!pending` check. Mutex
discipline is consistent — every `openGates`/`pendingUntil` access is
under `gatesMutex`. The fallback paths (auth failure, session expired
during delay) delete the map entry rather than leaving stale state.
This is careful work for code that runs unattended on routers.

### Blocking concern — `redirect_url` config is never delivered to the client

This is the functional gap that keeps the feature from doing what its
config schema promises. `main.go` reads `RedirectURL` but uses it only
as a boolean gate to enable `AuthDelay`
([main.go#L163-L178](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/main.go#L163-L178))
— the URL value itself is logged and then discarded. It is never passed
to the captive portal. Meanwhile `welcome.html` resolves its redirect
target from a hardcoded default (`wallet.cashu.me/welcome`) or an
untrusted `?redirect=` query parameter
([welcome.html#L89-L95](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/packaging/files/tollgate-captive-portal-site/welcome.html#L89-L95)):

```js
var TARGET = params.get('redirect') || DEFAULT_TARGET;
```

So an operator who sets `"redirect_url": "https://my-portal.example"` in
`config.json` gets the auth delay turned on, but their configured URL is
ignored — clients are still sent to the hardcoded default or whatever
`?redirect=` says. The config field's schema description ("URL to
redirect clients to after payment") is not implemented.

**Fix:** plumb `redirect_url` from config into the portal page. The
cleanest path is to have the backend inject it (template the value into
`welcome.html` at serve time, or expose it via an existing portal API
endpoint that the page fetches). Once that's in place, the `?redirect=`
query param should be removed or restricted to an allowlist — right now
it is a textbook open redirect on a captive-portal endpoint, where a
crafted link (`welcome.html?redirect=https://phishing.example`) sends a
paying user to an attacker-controlled site. This is the security item
worth addressing before this ships.

### Should-fix follow-ups (non-blocking)

- **No test coverage for the delayed-auth paths.** The existing
  `valve_test.go` covers the timer sentinel pattern, MAC validation, and
  `authorizeMAC` retry — but nothing exercises `AuthDelay > 0`,
  `delayedAuth`, `delayedAuthIndefinite`, the `pendingUntil` extend
  path, or the `CloseGate`-cancels-pending path. These are the most
  concurrency-sensitive additions in the PR. A test that sets
  `AuthDelay = 10ms`, stubs `runNdsctl`, calls `OpenGateUntil` twice
  during the window, then asserts the goroutine reads the extended
  timestamp would lock in the design.

- **`valve.Stop()` is dead code.**
  ([valve.go#L67-L69](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/valve/valve.go#L67-L69))
  It closes `stopCh`, but nothing in the codebase calls it — `main.go`
  has no signal handler or graceful-shutdown path. The goroutines'
  actual cancellation is via `CloseGate` deleting the pending entry (the
  `!pending` abort), not via `stopCh`. Either wire `Stop()` into a
  `signal.Notify` handler in `main()`, or remove it. As-is, it also
  carries a latent panic: `close(stopCh)` on an already-closed channel
  would crash, so if anyone ever calls it from two shutdown paths it
  blows up.

- **README config reference not updated.** The `config.json` example
  block
  ([README.md#L84-L138](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/README.md#L84-L138))
  and the "Key fields" list don't mention `redirect_url` or
  `auth_delay_seconds`. The CHANGELOG entry is good; the README should
  track it.

### Minor notes

- **`pendingUntil[mac] = 1` sentinel**
  ([valve.go#L263](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/valve/valve.go#L263))
  in `OpenGate`'s indefinite path uses unix-timestamp `1` as a marker.
  It works because `delayedAuthIndefinite` only checks map membership,
  not the value — but a separate `pendingIndefinite` set (or a bool
  alongside the timestamp) would be clearer to a future reader than a
  magic `1`.

- **Redundant double lookup in `OpenGate`**
  ([valve.go#L253-L256](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/src/valve/valve.go#L253-L256)):
  `_, exists := openGates[mac]` followed by
  `if existingTimer, ok := openGates[mac]; ok` does the same lookup
  twice. Collapsing to a single `existingTimer, exists :=` would be
  cleaner, though this is a skim-past.

- **`balance.html` hardcodes 3000ms**
  ([balance.html#L23-L27](https://github.com/OpenTollGate/tollgate-module-basic-go/blob/00b4a5d6367305e5b67ec352f4c1f3571df24a25/packaging/files/tollgate-captive-portal-site/balance.html#L23-L27))
  for the redirect to `welcome.html`, which itself counts down 5s —
  total ~8s, matching the default `AuthDelay`. But if an operator sets a
  different `auth_delay_seconds`, the HTML delays don't follow. Worth a
  comment documenting the intended alignment, or deriving the value from
  config once the plumbing lands.

- **PR body** is truncated mid-sentence ("resolved by merging upstream'"s")
  and doesn't describe the `welcome.html`/`balance.html` changes, the
  Android intent-URL handling, or the dual-delay architecture. Since the
  PR will be squash-merged with a maintainer-rewritten message this is
  cosmetic, but a complete body helps reviewers now.

- **Commit hygiene** — the three `fix:` commits are schema-placement
  fixups that would normally be squashed before merge. Non-issue given
  the squash-merge policy.

### Disposition: request-changes

The valve goroutine lifecycle is well-reasoned and I'd merge that part
today. But the headline feature — "configurable redirect URL" — doesn't
actually deliver the configured URL to the client, and the `?redirect=`
query param is an open redirect on a portal-facing endpoint. Both are
fixable in the HTML/plumbing layer without touching the valve logic.
I'd ask for:

1. Plumb `redirect_url` from config to `welcome.html` (template or API).
2. Remove or allowlist the `?redirect=` query param.
3. (Nice-to-have) a test for the delayed-auth extend/cancel paths.

Once those land, this is good to go.

c03rad0r (approved):
Reviewed. Solid implementation of post-payment UX delay.

Strengths:
- Goroutine lifecycle well-handled (cancellable via stopCh, cleanup in CloseGate, session expiry checked after delay)
- pendingUntil map preserves extensions during delay window — handles concurrent payment race
- Android Chrome intent URL pattern is the correct approach for captive portals
- Config fields (auth_delay_seconds, redirect_url) properly schema-declared

Minor notes for follow-up (not blocking):
1. welcome.html hardcodes https://wallet.cashu.me/welcome as default redirect — consider defaulting to no redirect or the gateway's own page
2. The 3-second setTimeout in balance.html is hardcoded, not tied to auth_delay_seconds config value
3. The ?redirect= query param in welcome.html is an open redirect — consider validating against an allowlist since this runs inside a captive portal

CI green. Approving.

Issue #207, opened 2026-07-11
Test: AP setup recovery on reinstall/upgrade (#103, #173)


Amperstrand:
## What to test

Issues #103 and #173 report that TollGate AP interfaces are not created when `/etc/tollgate-setup-done` already exists (reinstall/upgrade scenario). The `99-tollgate-setup` uci-defaults script exits entirely if the flag is found (line 8-9), skipping wireless AP creation.

## How to run

Requires a router reinstall test on the cloud lab or physical router:

```bash
# 1. Deploy TollGate fresh
ssh root@<router> 'opkg install tollgate-wrt'

# 2. Verify APs exist
ssh root@<router> 'iw dev | grep ssid'
# Expected: TollGate-* SSIDs on both radios

# 3. Simulate reinstall (setup-done flag persists)
ssh root@<router> 'opkg install tollgate-wrt --force-reinstall'

# 4. Reboot
ssh root@<router> 'reboot'

# 5. After reboot, verify APs STILL exist
ssh root@<router> 'iw dev | grep ssid'
# Bug: APs are MISSING because setup-done flag skips the script

# 6. Check the setup log
ssh root@<router> 'cat /tmp/tollgate-setup.log 2>/dev/null'
```

## Pass criteria (current/buggy state)

- Fresh install: APs created ✅
- Reinstall: APs missing ❌ (confirms the bug)

## Pass criteria (after fix)

- Fresh install: APs created ✅  
- Reinstall: APs still present ✅ (setup script verifies wireless config even when flag exists)

## Context

#103: original bug (78 days old). #173: the proposed fix is stale (NetworkMonitor path changed in PR #109). Root cause: line 8 of 99-tollgate-setup bails entirely when setup-done flag exists.

Amperstrand:
## Test Results — AP setup recovery (#103, #173)

**Status: Bug confirmed on main. ✅**

Tested on SHC baked Dev VPS (nested KVM, OpenWrt 24.10.1, main @ ff1fd12):

| Test | Result | Meaning |
|------|--------|---------|
| `test_setup_script_bails_on_existing_flag` | **PASSED** | Confirms the bug: `99-tollgate-setup` exits at line 8-9 when `/etc/tollgate-setup-done` exists, skipping `setup_wifi()` |
| `test_setup_wifi_step_exists` | **PASSED** | The AP-creation step exists in the script (lines 72-86) but is skipped by the early exit |
[...]

PR #216, opened 2026-07-11
fix: AP setup recovery on reinstall/upgrade (#103, #173)

### LINKED ISSUE #103: Bug: TollGate AP interfaces are not created when the setup flag already exists
LLM diagnosis below could be off. 

### Problem

When `tollgate-wrt` is installed on a router that already has the setup flag `/etc/tollgate-setup-done` (i.e., a reinstall or upgrade), the public `TollGate-*` access points do not appear even though internet is available. The `default_radio0` and `default_radio1` wifi-iface sections are never created.

On a fresh image install (no setup flag), the `99-tollgate-setup` uci-defaults script correctly creates these sections. The problem only manifests when the setup flag already exists and the AP sections are missing from `/etc/config/wireless`.

### Steps to reproduce

1. Flash a fresh image and let tollgate-wrt complete first-boot setup (creates APs and the setup flag)
2. Verify APs exist: `uci show wireless | grep default_radio` — should show `default_radio0` and `default_radio1`
3. Manually remove the AP sections (or install a different image that lacks them): `uci delete wireless.default_radio0; uci delete wireless.default_radio1; uci commit wireless; wifi reload`
4. Reinstall tollgate-wrt (or restart the service): `service tollgate-wrt restart`
5. Observe: no `TollGate-*` APs appear, even though internet is available

### Root Cause

When the setup flag exists, the `99-tollgate-setup` uci-defaults script exits immediately:

```sh
SETUP_FLAG="/etc/tollgate-setup-done"
if [ -f "$SETUP_FLAG" ]; then
    exit 0
fi
```

The only recovery path is the Go binary's `ensureAPInterfacesExist()` (`src/wireless_gateway_manager/connector.go:558`), which is called from `updatePriceAndAPSSID()` → `UpdateLocalAPSSID()`. However, `updatePriceAndAPSSID()` gates on the network monitor's connectivity check (`wireless_gateway_manager.go:279`):

```go
if !gm.networkMonitor.IsConnected() {
    logger.Info("Not connected to a gateway")
    return
}
```

The network monitor pings `8.8.8.8` every 30 seconds and requires at least 1 successful ping before reporting connected (`network_monitor.go:54-56`). When the service starts, the first ping hasn't completed yet, so `IsConnected()` returns `false` and `updatePriceAndAPSSID()` returns early — **permanently**. There is no retry mechanism.

This race condition occurs regardless of whether internet is available. The issue is that `updatePriceAndAPSSID()` fires before the network monitor has had time to confirm connectivity.

### Evidence

After `service tollgate-wrt restart` on a router with active STA connections and internet access, the logs show gateway probing and payout activity but **zero** log lines containing "Ensuring default AP interfaces exist" or "Not connected to a gateway", confirming `updatePriceAndAPSSID()` returned early and `ensureAPInterfacesExist()` was never reached.

### Proposed Fix

Add an `onConnected` callback to `NetworkMonitor` that fires once when connectivity is first confirmed by ping. `GatewayManager` registers a callback that calls `updatePriceAndAPSSID()`. This ensures AP interfaces are created once the network monitor confirms internet, regardless of whether connectivity was already available at service start time.

**`src/wireless_gateway_manager/types.go`** — add fields to `NetworkMonitor`:

```go
type NetworkMonitor struct {
    connector       *Connector
    pingFailures    int
    pingSuccesses   int
    isInSafeMode    bool
    ticker          *time.Ticker
    stopChan        chan struct{}
    onConnected     func()  // callback fired once when connectivity is established
    connectedOnce   bool    // guards against firing the callback multiple times
}
```

**`src/wireless_gateway_manager/network_monitor.go`** — fire callback on first successful ping:

```go
func (nm *NetworkMonitor) checkConnectivity() {
    err := ping(pingTarget)
    if err != nil {
        nm.pingFailures++
        nm.pingSuccesses = 0
        logger.WithField("consecutive_failures", nm.pingFailures).Warn("Ping failed")
    } else {
        nm.pingSuccesses++
        nm.pingFailures = 0
        logger.WithField("consecutive_successes", nm.pingSuccesses).Debug("Ping successful")

        if !nm.connectedOnce && nm.onConnected != nil {
            nm.connectedOnce = true
            logger.Info("Network connectivity established, calling onConnected callback")
            nm.onConnected()
        }
    }
}
```

**`src/wireless_gateway_manager/wireless_gateway_manager.go`** — register callback in `Init()`:

```go
// Before starting the network monitor:
networkMonitor.onConnected = func() {
    go gatewayManager.updatePriceAndAPSSID()
}

go gatewayManager.RunPeriodicScan(ctx)
gatewayManager.networkMonitor.Start()

// Still try immediately in case connectivity is already established
gatewayManager.updatePriceAndAPSSID()
```

The `go` before the callback is intentional — it fires from the network monitor's ping goroutine, and `updatePriceAndAPSSID()` calls `ExecuteUCI()` which may block.

### Files changed

- `src/wireless_gateway_manager/types.go`
- `src/wireless_gateway_manager/network_monitor.go`
- `src/wireless_gateway_manager/wireless_gateway_manager.go`

# Plan: Fix Missing TollGate AP Interfaces When Setup Flag Already Exists

## Problem

When `tollgate-wrt` is installed on a router that already has the setup flag `/etc/tollgate-setup-done` (i.e., a reinstall or upgrade), the public `TollGate-*` access points do not appear even though internet is available. The `default_radio0` and `default_radio1` wifi-iface sections are never created.

## Context

On a fresh image install (no setup flag), the `99-tollgate-setup` uci-defaults script correctly creates these sections. The problem only manifests when the setup flag already exists and the AP sections are missing from `/etc/config/wireless`.

## Root Cause

When the setup flag exists, the `99-tollgate-setup` uci-defaults script exits immediately:

```sh
SETUP_FLAG="/etc/tollgate-setup-done"
if [ -f "$SETUP_FLAG" ]; then
    exit 0
fi
```

The only recovery path is the Go binary's `ensureAPInterfacesExist()` (`src/wireless_gateway_manager/connector.go:558`), which is called from `updatePriceAndAPSSID()` → `UpdateLocalAPSSID()`. However, `updatePriceAndAPSSID()` gates on the network monitor's connectivity check (`wireless_gateway_manager.go:279`):

```go
if !gm.networkMonitor.IsConnected() {
    logger.Info("Not connected to a gateway")
    return
}
```

The network monitor pings `8.8.8.8` every 30 seconds and requires at least 1 successful ping before reporting connected (`network_monitor.go:54-56`). When the service starts, the first ping hasn't completed yet, so `IsConnected()` returns `false` and `updatePriceAndAPSSID()` returns early — **permanently**. There is no retry mechanism.

This race condition occurs regardless of whether internet is actually available. The issue is that `updatePriceAndAPSSID()` fires before the network monitor has had time to confirm connectivity via its 30-second ping interval.

### File: `src/wireless_gateway_manager/types.go`

Add two fields to `NetworkMonitor`:

```go
type NetworkMonitor struct {
    connector       *Connector
    pingFailures    int
    pingSuccesses   int
    isInSafeMode    bool
    ticker          *time.Ticker
    stopChan        chan struct{}
    onConnected     func()  // NEW: callback fired once when connectivity is established
    connectedOnce   bool    // NEW: guards against firing the callback multiple times
}
```

### File: `src/wireless_gateway_manager/network_monitor.go`

Modify `checkConnectivity()` to call the callback on the first successful ping:

```go
func (nm *NetworkMonitor) checkConnectivity() {
    err := ping(pingTarget)
    if err != nil {
        nm.pingFailures++
        nm.pingSuccesses = 0
        logger.WithField("consecutive_failures", nm.pingFailures).Warn("Ping failed")
    } else {
        nm.pingSuccesses++
        nm.pingFailures = 0
        logger.WithField("consecutive_successes", nm.pingSuccesses).Debug("Ping successful")

        // Fire onConnected callback once on first connectivity
        if !nm.connectedOnce && nm.onConnected != nil {
            nm.connectedOnce = true
            logger.Info("Network connectivity established, calling onConnected callback")
            nm.onConnected()
        }
    }
}
```

### File: `src/wireless_gateway_manager/wireless_gateway_manager.go`

In `Init()`, register the callback before starting the network monitor:

```go
func Init(ctx context.Context, configManager *config_manager.ConfigManager) (*GatewayManager, error) {
    connector := &Connector{}
    scanner := &Scanner{connector: connector}
    vendorProcessor := &VendorElementProcessor{connector: connector}
    networkMonitor := NewNetworkMonitor(connector)

    gatewayManager := &GatewayManager{
        scanner:           scanner,
        connector:         connector,
        vendorProcessor:   vendorProcessor,
        networkMonitor:    networkMonitor,
        configManager:     configManager,
        availableGateways: make(map[string]Gateway),
        currentHopCount:   math.MaxInt32,
        scanInterval:      30 * time.Second,
        stopChan:          make(chan struct{}),
    }

    // Register callback to create AP interfaces once connectivity is established.
    // This handles the race where updatePriceAndAPSSID() fires before the
    // network monitor confirms internet connectivity on first boot.
    networkMonitor.onConnected = func() {
        go gatewayManager.updatePriceAndAPSSID()
    }

    go gatewayManager.RunPeriodicScan(ctx)
    gatewayManager.networkMonitor.Start()

    // Still try immediately in case connectivity is already established
    gatewayManager.updatePriceAndAPSSID()

    return gatewayManager, nil
}
```

**Why `go` before the callback:** The callback fires from within the network monitor's ping goroutine. `updatePriceAndSSID()` calls `ExecuteUCI()` which may block. Using `go` prevents blocking the network monitor's ticker loop.

### No changes to interfaces.go

The `NetworkMonitorInterface` should remain unchanged. The callback is set directly on the concrete `NetworkMonitor` struct before `Start()` is called, not through the interface. This keeps the interface clean and the callback is an internal implementation detail.

### File: `src/wireless_gateway_manager/network_monitor_test.go` (new file)

Test the connectivity callback behavior. Use the existing mock patterns from `reseller_mode_test.go` (which already has `MockNetworkMonitor`):

```go
func TestNetworkMonitor_OnConnectedCallback_FiresOnceOnFirstPing(t *testing.T) {
    // Create a NetworkMonitor with a callback counter
    // Call checkConnectivity() with a simulated failed ping
    // Verify callback was NOT called
    // Call checkConnectivity() with a simulated successful ping
    // Verify callback WAS called exactly once
    // Call checkConnectivity() with another successful ping
    // Verify callback was NOT called again
}

func TestNetworkMonitor_OnConnectedCallback_NilCallbackSafe(t *testing.T) {
    // Create a NetworkMonitor without setting onConnected
    // Call checkConnectivity() with a successful ping
    // Verify no panic occurs
}
```

**Important:** `NetworkMonitor.checkConnectivity()` is unexported and calls the unexported `ping()` function. To test this, either:
- Export the test helper functions, or
- Test indirectly through the `Init()` public API (but that's heavier), or
- Move the ping logic to a `pingFunc` field on `NetworkMonitor` that can be swapped in tests (similar to how `onConnected` is set)

**Recommended approach:** Add a `pingFunc func(host string) error` field to `NetworkMonitor` that defaults to the real `ping()` but can be overridden in tests. Then test `checkConnectivity()` directly.

### Run tests

```bash
cd src/wireless_gateway_manager && go test -v ./...
```

## Manual verification on router

After deploying the fix, on a router where the setup flag exists but `default_radio0`/`default_radio1` are missing:

1. Restart the service:
   ```sh
   service tollgate-wrt restart
   ```

2. Wait ~30 seconds for the first successful ping, then check:
   ```sh
   logread | grep -i "ensuring default\|network connectivity established"
   uci show wireless | grep default_radio
   wifi status
   ```

3. Verify:
   - Log shows "Network connectivity established, calling onConnected callback"
   - Log shows "Ensuring default AP interfaces exist"
   - `default_radio0` and `default_radio1` exist in wireless config
   - `wifi status` shows the TollGate SSID broadcasting

## Summary of changes

| File | Change |
|------|--------|
| `src/wireless_gateway_manager/types.go` | Add `onConnected func()` and `connectedOnce bool` fields to `NetworkMonitor` |
| `src/wireless_gateway_manager/network_monitor.go` | In `checkConnectivity()`, call `onConnected` once on first successful ping |
| `src/wireless_gateway_manager/wireless_gateway_manager.go` | In `Init()`, register `onConnected` callback before starting the network monitor |
| `src/wireless_gateway_manager/network_monitor_test.go` | New test file for connectivity callback |

### LINKED ISSUE #173: AP setup is non-recoverable if sections are missing on reinstall/upgrade (real fix for #103)
## Problem

#103 reported that TollGate AP interfaces are not created when
`/etc/tollgate-setup-done` already exists. The originally-proposed fix is
**stale** — the `NetworkMonitor`-based Go recovery path it describes was
removed in PR #109 (WGM rewrite) and no longer exists on `main`.

**Current behavior:** if `/etc/tollgate-setup-done` exists and
`/etc/config/wireless` is missing `default_radio0` / `default_radio1`, there
is **no code path** — shell or Go — that recreates them. The `TollGate-*`
APs never appear until an operator manually removes the setup flag and
reboots.

This is the HIGH-risk upgrade scenario called out in the v0.5.0 release plan
(#154): UCI-defaults were rewritten in #84 / #90 / #96, so a real upgrade
landing in this state is plausible.

## Repro

```sh
# Router with TollGate APs running and setup flag present.
uci delete wireless.default_radio0
uci delete wireless.default_radio1
uci commit wireless
wifi reload
service tollgate-wrt restart      # or just reboot
# Wait 5 minutes.
uci show wireless | grep default_radio   # → empty (BUG)
wifi status | grep -i ssid              # → no TollGate SSID (BUG)
```

## Root cause

`packaging/files/etc/uci-defaults/99-tollgate-setup` does:

```sh
if [ -f "$SETUP_FLAG" ]; then
    exit 0
fi
```

…before any AP setup. The only place AP sections are created is inside
`setup_public_wifi()` → `configure_radio_ap()` after that early-exit gate. So
when the flag exists, AP setup is unreachable.

## Proposed fix: Option A (idempotent uci-defaults, recommended for v0.5.0)

Smallest possible change. The script already runs on every boot; just let it
repair the AP sections when they're missing, without re-running the
first-boot-only steps.

```sh
SETUP_FLAG="/etc/tollgate-setup-done"

if [ -f "$SETUP_FLAG" ]; then
    # Setup has run, but AP sections might be missing (reinstall, partial
    # wipe, manual `uci delete`). Repair them and bail — don't re-run the
    # one-shot steps that the flag guards.
    if ! uci -q get wireless.default_radio0 >/dev/null 2>&1 || \
       ! uci -q get wireless.default_radio1 >/dev/null 2>&1; then
        log "Setup flag exists but AP sections missing — repairing"
        RANDOM_SUFFIX=$(hexdump -n 3 -e '4/1 "%02X"' /dev/urandom | cut -c1-4)
        GATEWAY_NAME="TollGate-${RANDOM_SUFFIX}"
        configure_radio_ap 'default_radio0' 'tollgate_2g_open'
        configure_radio_ap 'default_radio1' 'tollgate_5g_open'
        uci -q get wireless.radio0 >/dev/null && uci set wireless.radio0.disabled='0'
        uci -q get wireless.radio1 >/dev/null && uci set wireless.radio1.disabled='0'
        uci commit wireless
        wifi reload
    fi
    exit 0
fi

# ...rest of first-boot flow unchanged
```

### Why not regenerate the same SSID?

The original SSID had a random suffix (`TollGate-${RANDOM_SUFFIX}`). Clients
that saved it won't auto-reconnect to a new suffix. We could persist the
suffix to a file (e.g. `/etc/tollgate/gateway-name`) and re-use it on repair;
that's a follow-up nicety, not required for v0.5.0.

### Considerations / edge cases

- STA-mode `default_radio*` sections must not be clobbered. The existing
  `configure_radio_ap()` already guards with `current_mode == "sta"` → early
  return, so this is safe.
- If only one of the two sections is missing, only that one is repaired — the
  guard checks each independently.
- `wifi reload` after `uci commit wireless` is what makes this heal on the
  next reboot without a separate service restart.

## Alternative considered: Go-side idempotent setup at service start (Option B)

A second option would add `EnsureAPInterfaces()` to the Go service entrypoint
that mirrors the shell logic. Pros: survives even if uci-defaults are
skipped. Cons: re-introduces a Go path that has to stay in sync with the
shell, and the uci-defaults path already runs on every boot. Not recommended
for v0.5.0 — revisit only if Option A turns out to be insufficient in
practice.

## Acceptance criteria

- [ ] On a router with `setup-done` flag set and `default_radio0/1` missing,
      restarting the service **or** rebooting causes the APs to reappear
      within ~60 s.
- [ ] STA-mode sections are preserved (existing regression guard).
- [ ] Hardware regression test on `physical-router-test-automation` passes
      (new method in `test_tag_readiness.py`).
- [ ] No regression in the first-boot flow (existing tag-readiness preflight
      still passes — both `default_radio0` and `default_radio1` broadcast).

## Related

- #103 — original bug report. The proposed fix there is stale; this issue
  replaces it with the actual current-architecture fix.
- #154 — release plan, "Upgrade from v0.4.0 — HIGH" risk line.
- PR #109 — WGM rewrite that removed the Go-side recovery path.
- Companion hardware test (link TBD on `physical-router-test-automation`).

Amperstrand:
## Problem

The `99-tollgate-setup` uci-defaults script bailed entirely (`exit 0`) when `/etc/tollgate-setup-done` existed. On reinstall/upgrade, the flag persists in `/etc/`, so wireless APs were never recreated — TollGate SSIDs disappeared.

## Fix

Both approaches combined:

1. **Version-gated flag**: the flag now stores the setup version (`v0.5.0`). Upgrades that ship a new version trigger full re-setup.
2. **Always-run wireless verification**: even on same-version reinstall, the script verifies TollGate APs exist on both radios and recreates them if missing. AP SSID is recovered from existing config for consistency.

## Before (buggy)
```sh
if [ -f "$SETUP_FLAG" ]; then exit 0; fi  # bails entirely
```

## After (fixed)
```sh
if [ "$EXISTING_VERSION" = "$SETUP_VERSION" ]; then
    # Same version: verify/recreate APs only, then exit
    configure_radio_ap 'default_radio0' 'tollgate_2g_open'
    configure_radio_ap 'default_radio1' 'tollgate_5g_open'
    uci commit wireless
    exit 0
fi
# Full setup for new version or first boot
echo "$SETUP_VERSION" > "$SETUP_FLAG"
```

## Test plan

See #207 for manual reinstall test commands. Cloud-lab test: deploy → verify APs → reinstall → verify APs persist.

Fixes #103, fixes #173.

### DIFF
--- a/packaging/files/etc/uci-defaults/99-tollgate-setup
+++ b/packaging/files/etc/uci-defaults/99-tollgate-setup
@@ -2,12 +2,13 @@
 # TollGate first-boot configuration. Runs once from /etc/init.d/boot before
 # services start. All UCI changes are committed in a single pass at the end;
 # no service restarts here (uci-defaults runs before procd brings services up).
+#
+# The flag file stores the setup version. On reinstall/upgrade:
+# - Same version → verify wireless APs only (belt-and-suspenders)
+# - New version or first boot → full setup
 
 SETUP_FLAG="/etc/tollgate-setup-done"
-
-if [ -f "$SETUP_FLAG" ]; then
-    exit 0
-fi
+SETUP_VERSION="v0.5.0"
 
 LOGFILE=/tmp/tollgate-setup.log
 log() { echo "$(date '+%Y-%m-%d %H:%M:%S') - $1" >> "$LOGFILE"; }
@@ -277,8 +278,36 @@ commit_all() {
 
 # -- driver -----------------------------------------------------------------
 
+# Check if full setup is needed (version mismatch or first boot)
+EXISTING_VERSION=""
+if [ -f "$SETUP_FLAG" ]; then
+    EXISTING_VERSION=$(cat "$SETUP_FLAG" 2>/dev/null)
+fi
+
+if [ "$EXISTING_VERSION" = "$SETUP_VERSION" ]; then
+    # Same version — verify wireless APs only (fixes #103/#173: APs missing
+    # after reinstall when setup-done flag persisted from a prior install).
+    : > "$LOGFILE"
+    log "Flag matches $SETUP_VERSION — verifying wireless APs only"
+
+    # Recover gateway name from existing config for AP SSID consistency
+    GATEWAY_NAME=$(uci -q get wireless.tollgate_2g_open.ssid 2>/dev/null) || \
+    GATEWAY_NAME=$(uci -q get wireless.default_radio0.ssid 2>/dev/null) || \
+    GATEWAY_NAME="TollGate-$(hexdump -n 3 -e '4/1 "%02X"' /dev/urandom | cut -c1-4)"
+    RANDOM_SUFFIX="${GATEWAY_NAME#TollGate-}"
+
+    configure_radio_ap 'default_radio0' 'tollgate_2g_open'
+    configure_radio_ap 'default_radio1' 'tollgate_5g_open'
+    uci -q get wireless.radio0 >/dev/null && uci set wireless.radio0.disabled='0'
+    uci -q get wireless.radio1 >/dev/null && uci set wireless.radio1.disabled='0'
+    uci commit wireless
+    log "Wireless AP verification completed"
+    exit 0
+fi
+
+# Full setup (first boot or version change)
 : > "$LOGFILE"
-log "Starting TollGate configuration"
+log "Running full setup (version $SETUP_VERSION, previous: ${EXISTING_VERSION:-none})"
 
 setup_firewall_include
 setup_uhttpd
@@ -293,6 +322,6 @@ enable_modems
 setup_private_network        # consumes RANDOM_SUFFIX
 commit_all
 
-log "TollGate setup completed"
-touch "$SETUP_FLAG"
+log "TollGate setup completed (version $SETUP_VERSION)"
+echo "$SETUP_VERSION" > "$SETUP_FLAG"
 exit 0


Amperstrand:
## Cloud lab verification — SHC, full API suite

Tested on SHC Dev VPS (nested KVM, OpenWrt 24.10.1, x86_64, local mints):

```
171 passed, 0 failed, 0 errors, 144 skipped
```

**Zero regressions** — no existing API test broke. The setup script now uses SETUP_VERSION-aware logic (verified on the deployed OpenWrt VM). The unconditional `exit 0` on the setup-done flag is replaced with version-aware AP verification.

c03rad0r (approved):
## Review — AP Setup Recovery on Reinstall/Upgrade

**Verdict: Approved.** Small, targeted fix for two real operational bugs.

### What this fixes
When you reinstall or upgrade the tollgate package, the uci-defaults script (`99-tollgate-setup`) checks if `/etc/tollgate-setup-done` exists. If it does, it bails entirely — which means wireless APs never get recreated and disappear on upgrade.

The fix: (1) stores a version string (`v0.5.0`) in the flag file, so a new version triggers full re-setup even if the flag exists; (2) always runs wireless verification even on same-version reinstall.

### Pros
- Small (+36/-7), defensive — verifies AP sections exist even when flag says "done".
- Fixes #103 and #173.

### Minor note (non-blocking)
- Version comparison is string-based (could break on non-semver versions) but acceptable for this use case.

🚀

c03rad0r (approved):
## Review: AP Setup Recovery on Reinstall/Upgrade

**Verdict: Approved** ✅

Fixes #103, #173. Small change: +36/-7 lines, one shell script (`uci-defaults`).

### What it does

When you reinstall or upgrade the tollgate package, the `99-tollgate-setup` uci-defaults script checks if `/etc/tollgate-setup-done` exists. If it does, it bails entirely (`exit 0`) — which means wireless APs never get recreated and disappear on upgrade. This PR fixes it by:

1. **Version-gated flag**: stores a version string (`v0.5.0`) in the flag file, so a new version triggers full re-setup even if the flag exists.
2. **Always-run wireless verification**: even on same-version reinstall, the script verifies TollGate APs exist on both radios and recreates them if missing.

### Why it's needed

Real operational bug. Users upgrading their firmware lose their AP interfaces. Broken out-of-box experience.

### Pros

- Small, targeted, fixes two real issues (#103, #173).
- CI fully green.
- Defensive — verifies AP sections exist even when flag says "done".
- Recovers AP SSID from existing config for consistency.

### Cons

None significant. Version comparison is string-based (could break on non-semver versions) but acceptable for this use case.

Quick win, ready to merge.

PR #191, opened 2026-07-08
feat(identity): switch to 12-word NIP-06 mnemonic (from 24-word raw-entropy)

Amperstrand:
## What

Replace the non-standard "key-as-BIP39-entropy" encoding (24 words) with standard NIP-06 hierarchical derivation (12 words, 128 bits).

Depends on PR #189 merging first — this branch builds on top of it.

## Why

- secp256k1 provides ~128 bits of security (Pollard's rho on the 256-bit group). 12 words = 128 bits = **full curve security**. 24 words added zero security, only backup burden.
- NIP-06 (`m/44'/1237'/0'/0/0`) is the standard Nostr key derivation from a BIP39 seed, supported by `nak`, nos2x, and other clients.
- The previous approach (using the 32-byte private key directly as BIP39 entropy → 24 words) was non-standard and not compatible with any existing Nostr wallet's mnemonic import.

## Changes

| Function | Before | After |
|---|---|---|
| `GenerateMnemonic()` | — | generates 12-word mnemonic (128-bit entropy) |
| `MnemonicToPrivateKey()` | raw-entropy decode (24 words → key) | NIP-06 derivation via `m/44'/1237'/0'/0/0` |
| `GenerateIdentity()` | — | mnemonic + key + all derived attrs |
| `DeriveFromMnemonic()` | — | recovery from mnemonic |
| `PrivateKeyToMnemonic()` | key → 24 words | **removed** (can't reverse NIP-06) |
| `RevealSeed()` | key → full identity | **removed** (replaced by GenerateIdentity / DeriveFromMnemonic) |
| Handler `POST /identity/reveal-seed` | takes key, returns mnemonic | takes mnemonic in POST body, returns full identity |

## Tests

- 19 identity package tests pass
- 5 handler tests pass
- SHC VM test suite results will be posted as a follow-up comment

## Depends on

- PR #189 (identity package) must merge first

PR #232, opened 2026-07-13
feat(identity): switch to 12-word NIP-06 mnemonic (replaces #191)

Amperstrand:
Replaces #191 (closed to enable maintainerCanModify).

**Allow edits by maintainers is enabled** — @c03rad0r can push improvements directly to this branch.

Original description from #191: Switches identity mnemonic from 24-word raw-entropy to 12-word NIP-06 mnemonic derivation. Related to #231 (first-boot handover).


c03rad0r (changes requested):
Reviewed — the NIP-06 derivation logic is correct and the crypto is sound. The `m/44'/1237'/0'/0/0` path implementation using nbd-wtf/go-nostr/nip06 is properly done, and 128 bits of entropy (12 words) is sufficient given secp256k1's ~128-bit security ceiling. Test coverage is good.

However, the following need to be resolved before merge:

---

**Required changes:**

- [ ] **1. Conflicts with PR #231 (first-boot secret handover).** This PR uses 12-word NIP-06 derivation; #231 uses 24-word raw-entropy mnemonic + NATO word passwords. Both add a `src/identity/` module with incompatible APIs. They cannot both merge as-is. These PRs need to be reconciled into a single coherent identity approach before either can land. Recommendation: use the NIP-06 12-word derivation from this PR as the base, and port the password derivation + handover mechanisms from #231 on top of it.

- [ ] **2. Replace FactomProject/basen dependency.** FactomProject is a defunct blockchain project with no active maintenance. The basen library is just base58 encoding — use `btcsuite/btcutil/base58` instead, which is actively maintained and widely used in the Go Bitcoin/Lightning ecosystem.

- [ ] **3. Migration path for existing routers.** Routers already provisioned with 24-word raw-entropy mnemonics have no upgrade path in this PR. If any routers are deployed, they would need re-provisioning. At minimum, document the breaking change in the PR description and CHANGELOG. Ideally, add a detection + migration path (detect 24-word mnemonic on boot, re-derive using old method, write new NIP-06 derived key).

---

**Note on #231 approval:** #231 has been approved with the understanding that IPv4/MAC derivation from the public key is correct (these are not secrets). The approval includes a suggestion for future BIP85-style rotatable identity derivation. The reconciliation between these two PRs should preserve that design decision.

c03rad0r:
Cross-reference: #231 and #232 must be treated as a **PAIR**. They both add incompatible `src/identity/` modules — #232 uses 12-word NIP-06 derivation (`m/44'/1237'/0'/0/0`), #231 uses 24-word raw-entropy mnemonic + password derivation + handover endpoints. Neither can merge independently. They must be reconciled into a single PR that combines the NIP-06 derivation approach from #232 with the password derivation and secret handover mechanisms from #231. See change-request checklists on both PRs.

Amperstrand:
Closing in favor of unified identity PR. NIP-06 derivation is preserved as the base; password derivation + RevealSeed from #231 are added on top.

PR #237, opened 2026-07-14
fix(packaging): remove dangling firewall-tollgate references from Makefile

c03rad0r:
## Bug

PR #196 deleted `files/etc/config/firewall-tollgate` (it was a dead iptables-syntax include that fw4 silently rejected), but two Makefile references were left behind:

1. **Install rule** (line ~164): `$(INSTALL_DATA) .../firewall-tollgate $(1)/etc/config/`
2. **Conffiles list** (line ~194): `/etc/config/firewall-tollgate`

This breaks **every** package build job on every architecture:

```
install: cannot stat '/builder/package/tollgate-wrt/files/etc/config/firewall-tollgate': No such file or directory
make[2]: *** [Makefile:215: ...tollgate-wrt.installed] Error 1
ERROR: package/tollgate-wrt failed to build.
```

## Fix

Remove both references. The firewall rules that #196 replaced them with are created directly via UCI named sections in `99-tollgate-setup`, so the install step and conffiles entry are no longer needed.

## Testing

CI on this PR should go fully green — all 6 architectures x ipk/apk variants.

## Files changed

- `packaging/Makefile` — 2 lines removed (install rule + conffiles entry)
- `CHANGELOG.md` — entry added under Fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and [Semantic Versioning](https://semver.org/).
   was silently rejected by fw4 (nftables); rules now created directly via
   UCI named sections with idempotent guards
   ([#196](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/196)).
+- **Makefile references to deleted firewall-tollgate.** PR #196 removed
+  `files/etc/config/firewall-tollgate` but two Makefile references (install
+  rule + conffiles list) were left behind, breaking all package builds
+  ([#235](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/235)).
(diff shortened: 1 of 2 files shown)


Amperstrand (approved):
## Approved — critical fix, unblocks all package builds

#196 deleted `files/etc/config/firewall-tollgate` but left two Makefile references:
1. `$(INSTALL_DATA) ... firewall-tollgate ...` in the install rule
2. `/etc/config/firewall-tollgate` in the conffiles list

Both reference a file that no longer exists → `opkg`/`apk` build fails. This 3-line deletion fixes it.

Verified the diff: only removes the two dangling references. No other changes. Merge immediately — main CI is broken without this.

PR #238, opened 2026-07-14
fix(ci): remove remaining firewall-tollgate references from build workflow + keep list

Amperstrand:
PR #237 fixed the Makefile but missed two references that cause every build to fail:

1. **.github/workflows/build-package.yml:302** — inline `install` command copying the deleted file
2. **packaging/files/lib/upgrade/keep.d/tollgate:2** — sysupgrade keep list entry

Error on every architecture:
```
install: cannot stat 'packaging/files/etc/config/firewall-tollgate': No such file or directory
```

This blocks all CI artifact production, which blocks the physical-router-test-automation smoke tests.


Amperstrand:
This completes #237 — catches the two references the Makefile fix missed:
1. `build-package.yml:302` — inline `install` command (this is the line actually failing the build)
2. `keep.d/tollgate:2` — sysupgrade keep list

Same pattern as #237, just different files. Build has been failing since #196 merged.

c03rad0r (approved):
Correct fix — removes two dead references to a file deleted in #196/#237. CI package builds have been broken since. This unblocks all artifact production.

c03rad0r (approved):
LGTM. Confirmed the referenced file is deleted (404). Two dead references removed, zero additions. This unblocks all package CI.

Issue #239, opened 2026-07-14
Reseller Bootstrap: Zero-Float Mesh Formation (Implementation Plan)


c03rad0r:
# Reseller Bootstrap: Zero-Float Mesh Formation

## Problem

A freshly configured TollGate in reseller mode has no internet, no upstream
session, and no ecash balance. When a customer pays, the reseller cannot
grant internet access because it has none to grant. The network cannot
bootstrap without pre-funding or trust.

## Solution

The first customer's first proof is forwarded in full to the upstream
TollGate. No split, no swap, no margin on the first transaction. The
upstream authorizes the reseller's STA MAC, granting a data session. The
reseller now has internet. Every subsequent payment earns margin normally.

This works recursively — a reseller-of-a-reseller bootstraps the same way.
Each hop adds latency only on the first transaction (5-15s for payment
forwarding). After that, the reseller can reach the mint directly for
proof swaps.

No walled gardens. No whitelists. No pre-funding. No trust. The first
proof is the spark that lights the network.

### First customer transaction (cold start)

1. Customer connects to reseller WiFi, sees captive portal, pays 2 sats
2. Reseller detects: no active upstream session → bootstrap mode
3. Reseller forwards the ENTIRE proof to upstream via Nostr payment
4. Upstream receives proof, validates at mint, authorizes reseller STA MAC
5. Reseller's upstream session begins (2 MB at 1 sat/MB upstream rate)
6. Reseller opens valve for customer: 1 MB (what 2 sats buys at reseller rate)
7. Reseller retains 1 MB of prepaid upstream bandwidth as float

**Customer UX during cold start (5-15s):**
- "Payment received"
- "Establishing internet link..."
- "Link established — N hops to gateway"
- "Access granted!"

### Second transaction (warm path)

1. Customer pays 2 sats (renewal or new customer)
2. Reseller HAS internet — reaches mint directly
3. Reseller swaps proof at mint, splits: 1 sat upstream + 1 sat profit
4. Sends 1 sat to upstream for 1 MB more
5. Customer gets 1 MB

Every subsequent transaction: normal margin. The "cost" of cold start
was zero — the reseller didn't lose money, it just didn't earn margin
on transaction #1, and gained 1 MB of prepaid float.

### Recursive multi-hop

A chain of TollGates A → B → C:

1. Customer pays C (hop 2 reseller)
2. C has no internet → forwards proof to B
3. B has no internet → forwards proof to A
[...]

PR #240, opened 2026-07-14
fix: protocol compliance — map notice codes to TIP-01 spec, fix tips tag

c03rad0r:
## Summary

Two protocol-compliance fixes bundled per maintainer request.

### 1. Notice event codes mapped to TIP-01 spec

The spec (TIP-01) defines four error codes for kind 21023 notice events:
- `session-error`
- `upstream-error-not-connected`
- `payment-error-mint-not-accepted`
- `payment-error-token-spent`

The implementation used its own non-spec codes. This PR maps them:

| Before | After | Rationale |
|--------|-------|-----------|
| `session-management-failed` | `session-error` | Spec code, detail in content |
| `gate-open-failed` | `session-error` | Gate failure is session management |
| `allotment-calculation-failed` | `session-error` | Part of session creation |
| `payment-error-token-spent` | (unchanged) | Already matches spec ✓ |
| `payment-error-invalid-token` | (unchanged) | No spec equivalent — kept |
| `invalid-mac-address` | (unchanged) | Pre-payment validation |
| `payment-processing-timeout` | (unchanged) | No spec equivalent |
| `payment-processing-failed` | (unchanged) | No spec equivalent |

Codes with no spec equivalent are kept as-is — the precision goes in the human-readable content string, following how general-purpose protocols work (e.g., HTTP status codes are general, error messages are specific).

### 2. Tips tag fixed

Advertisement event claimed `["tips", "1", "2", "3", "4"]` but only TIP-01 and TIP-02 exist. Changed to `["tips", "1", "2"]`.

## Testing

- `go build ./...` ✓
- `go vet ./...` ✓
- `go test -race -count=1 -tags testenv ./...` ✓ (all pass)
- `gofmt` clean on changed files

## Related

- Spec PR: OpenTollGate/tollgate#8 (TIP-01 harmonization)
- Issue: #239 (reseller bootstrap plan)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,17 @@ and [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **Protocol compliance: notice event codes and tips tag.** Map
+  implementation-specific notice event codes to spec-defined codes from
+  TIP-01 (`session-management-failed`, `gate-open-failed`, and
+  `allotment-calculation-failed` → `session-error`;
+  `payment-error-token-spent` already matched). Codes with no spec
+  equivalent (`payment-error-invalid-token`, `invalid-mac-address`,
+  `payment-processing-timeout`, `payment-processing-failed`) are kept
+  as-is with precision in the content string. Also remove non-existent
+  TIP-03 and TIP-04 from the advertisement `tips` tag — only TIP-01 and
+  TIP-02 are defined.
+
 - **Wireless config missing-file guard.** `scanner.GetRadios()` and
   `connector.getRadiosFromConfig()` now return gracefully when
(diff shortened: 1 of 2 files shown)


Amperstrand (approved):
## Approved — protocol compliance fix

**Notice codes:** Correct mapping to TIP-01 spec codes. The implementation-specific codes (`session-management-failed`, `gate-open-failed`, `allotment-calculation-failed`) → `session-error`. Codes with no spec equivalent kept as-is with detail in content string. Correct.

**Tips tag:** Removed non-existent TIP-03 and TIP-04 from the advertisement `tips` tag. Only TIP-01 and TIP-02 are defined in the spec. Correct.

Clean, well-documented change. CHANGELOG entry included.

PR #264, opened 2026-07-18
feat: add POST /retry-auth endpoint (issue #258)

### LINKED ISSUE #258: Payment succeeds but gate-open fails — user loses token with no refund
## Bug

When Cashu payment processing succeeds but `ndsctl auth` (gate opening) fails, the user loses their token without gaining internet access. There is no refund or retry mechanism.

### Reproduction (SHC virtual lab)

```
POST / with fresh Cashu token
→ PurchaseSession: Receive completed, amount=1, err=<nil> ← PAYMENT OK
→ Amount after swap: 1 ← SWAP OK
→ Error authorizing MAC address: exit status 1 ← NDS AUTH FAILS
→ Response: kind=21023, code=gate-open-failed
→ Session: session_active=false, remaining=0 ← NO SESSION CREATED
```

### Impact

**The user's Cashu token is irrevocably spent** but they have no internet access. There is no way to:
1. Retry the gate opening without re-paying
2. Get a refund of the spent token
3. Recover the lost session

### Suggested fix

Option A: **Delay payment processing** — open the gate first, then process payment. If gate fails, return error without spending the token.

Option B: **Add a retry endpoint** — `POST /retry-auth?mac=<mac>` that re-attempts `ndsctl auth` for a recently-paid session without requiring a new token.

Option C: **Track pending sessions** — store the session even when gate-open fails, and allow the user to retry via `GET /usage` (which would show the pending session).

### Related
- Discovered during SHC testing on v0.5.0
- The NDS auth failure is specific to the virtual lab (bridge MAC not recognized), but the no-refund behavior affects all deployments
- Concurrent payments from the same mint also show this: one succeeds, one fails, both tokens are spent

Amperstrand:
## What

When Cashu payment succeeds but NDS gate-open fails, users previously lost their token with no recovery. This PR adds POST /retry-auth which re-attempts ndsctl auth without requiring a new Cashu token.

## How

1. On gate-open failure: grantSessionAccess records the paid allotment via recordPendingAuth before rolling back the session
2. POST /retry-auth (empty body): MAC derived from source IP, looks up pending record, re-invokes grantSessionAccess
3. Abuse prevention: max 5 attempts per MAC within 10-minute window, no pending record means zero ndsctl calls
4. No refunds per design decision: small granular payments make refunds unnecessary

## Files

- src/merchant/pending_auth.go (new, 108 lines) - record type, RetryAuth, helpers
- src/merchant/merchant.go - struct fields, interface method, init
- src/merchant/merchant_degraded.go - stub
- src/merchant/lightning.go - grantSessionAccess hooks (2 lines)
- src/merchant/merchant_provider_test.go - mock update
- src/main.go - HandleRetryAuth handler + route

## Testing

- Build: pass
- All merchant tests with -race: pass (4.6s)
- Design reviewed by Oracle

Closes #258

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -472,6 +472,46 @@ func HandleRootPost(w http.ResponseWriter, r *http.Request) {
 
 }
 
+func HandleRetryAuth(w http.ResponseWriter, r *http.Request) {
+	mainLogger.WithFields(logrus.Fields{
+		"method":      r.Method,
+		"remote_addr": r.RemoteAddr,
+	}).Info("Received retry-auth request")
+
+	if r.Method != http.MethodPost {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	ip := getIP(r)
+	macAddress, err := getMacAddress(ip)
+	if err != nil {
+		mainLogger.WithError(err).Error("retry-auth: MAC lookup failed")
+		sendNoticeResponse(w, merchantProvider.inner.GetMerchant(), http.StatusBadRequest, "error", "mac-address-lookup-failed",
+			fmt.Sprintf("Failed to lookup MAC address for IP %s: %v", ip, err), "")
+		return
+	}
+
+	r.Body.Close()
+
+	responseEvent, err := merchantProvider.inner.GetMerchant().RetryAuth(macAddress)
+	w.Header().Set("Content-Type", "application/json")
+
+	if err != nil {
+		mainLogger.WithError(err).Error("retry-auth: internal error")
+		sendNoticeResponse(w, merchantProvider.inner.GetMerchant(), http.StatusInternalServerError, "error", "internal-error",
+			fmt.Sprintf("Internal error during retry-auth: %v", err), macAddress)
+		return
+	}
+
+	if responseEvent.Kind == 21023 {
+		w.WriteHeader(http.StatusBadRequest)
+	} else {
+		w.WriteHeader(http.StatusOK)
+	}
+	json.NewEncoder(w).Encode(responseEvent)
+}
+
 // sendNoticeResponse creates and sends a notice event response
 func sendNoticeResponse(w http.ResponseWriter, m merchant.MerchantInterface, statusCode int, level, code, message, customerPubkey string) {
 	noticeEvent, err := m.CreateNoticeEvent(level, code, message, customerPubkey)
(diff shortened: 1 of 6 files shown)


Amperstrand:
Closing — the payment-gate race is an accepted architectural limitation of the nodogsplash (NDS) model, not a bug to fix with an endpoint.

## Why this race exists

The Go v1 uses NoDogSplash for captive portal enforcement. Payment processing and gate control are **separate processes** communicating via `ndsctl auth` IPC. When payment succeeds but `ndsctl` fails (race condition, ndsctl timeout, NDS process issue), the user loses their token. This is inherent to the MAC-based NDS architecture — it cannot be fixed by adding a retry endpoint because the retry would face the same IPC boundary.

## Why a retry-auth endpoint is the wrong fix

A `POST /retry-auth` endpoint patches the symptom while adding:
- New API surface and client-side state management
- Ambiguity about who triggers the retry (portal JS? backend? background?)
- New edge cases (expired sessions, permanently broken ndsctl, concurrent retries)

The existing mitigations already handle transient ndsctl failures:
1. **Valve retry** (`valve.go`): 5 attempts with 400ms delay — handles the common case
2. **Token recovery path**: self-heals within 60-90s
3. **Lightning monitor backoff** (PR #249): retries `ensureLightningAccessGranted` on ndsctl failures

## What the Rust implementation learned (tollgate-rs)

Research in [tollgate-rs](https://github.com/Amperstrand/tollgate-rs) confirms the race is architectural, not implementational. Three active threads converge on eliminating it:

### 1. Drop nodogsplash entirely — [tollgate-rs#67](https://github.com/Amperstrand/tollgate-rs/issues/67)
Replace NDS with direct nftables rules managed by the same process:
```
Payment succeeds → backend adds IP to nftables paid_peers set → gate opens
```
Same process, no IPC, atomic payment+gate. The race **cannot exist** because there is no inter-process boundary. The Rust implementation is actively pursuing this.

### 2. RADIUS-based ecash auth — [tollgate-rs#51](https://github.com/Amperstrand/tollgate-rs/issues/51)
Prototype where the Cashu token IS the RADIUS credential. Payment and access are the **same atomic operation** — if payment fails, RADIUS auth fails, user never gets on the network. No race because there is no separate "gate" step.

### 3. NUT-18/NUT-24 wallet-native payments — [tollgate-rs#68](https://github.com/Amperstrand/tollgate-rs/issues/68)
Wallets handle payment natively via HTTP 402. The portal-mediated copy-paste-token model (and its race condition) is eliminated entirely.

## Decision

For **internet tollgates**, payment-first is the correct architecture — internet access cannot be un-delivered. The accepted risk window (payment succeeds, gate-open fails) is transient and self-healing.

For **physical services** (fuel dispensers, vending machines), a start-service-then-claim model might make sense — but that requires a fundamentally different architecture than what NDS provides.

The correct fix is the architectural migration away from NDS (tollgate-rs#67), not a retry endpoint on the NDS model. The Go v1 stays as-is with accepted risk.

Issue #263, opened 2026-07-18
refactor: create sysexec package — fix 11 fire-and-forget exec.Command calls + add context/timeout/logging/retry


Amperstrand:
## Problem: 11 fire-and-forget exec.Command calls + 26 without context/timeout

The codebase has **37 `exec.Command` calls across 7 files**. Only **2 use `exec.CommandContext`** (valve.go). The remaining 35 calls lack timeouts, context cancellation, and consistent error handling.

### Critical: Fire-and-forget calls (errors completely ignored)

These 11 calls execute **critical network operations** and silently discard errors:

```go
// wireless_gateway_manager/connector.go — after upstream switch
exec.Command("/etc/init.d/dnsmasq", "restart").Start()   // line 966
exec.Command("/etc/init.d/firewall", "restart").Start()   // line 967
exec.Command("ifdown", "wwan").Run()                      // line 976
exec.Command("ifup", "wwan").Run()                        // lines 998, 1075, 1079
exec.Command("/etc/init.d/network", "reload").Run()       // line 1150
exec.Command("/etc/init.d/firewall", "reload").Run()      // line 1192
exec.Command("wifi", "up").Run()                          // line 1231

// wireless_gateway_manager/upstream_manager.go
exec.Command("ifup", "wwan").Run()                        // line 147
```

If `ifup wwan` fails, the router loses upstream connectivity with **no log entry, no retry, no alert**. Operators cannot diagnose the issue.

### Existing gold standard: valve/valve.go

The valve package already implements the correct pattern:

```go
var runNdsctl = func(args ...string) (string, error) {
    ctx, cancel := context.WithTimeout(context.Background(), ndsctlTimeout)
    defer cancel()
    cmd := exec.CommandContext(ctx, "ndsctl", args...)
    output, err := cmd.CombinedOutput()
    return string(output), err
}
```

Plus: `isValidMAC()` input validation, `authMaxAttempts` retry logic, `stopCh` graceful shutdown, `logrus` structured logging.

### Design

```go
package sysexec

// Runner executes system commands with context, logging, and retry support.
type Runner interface {
    Run(ctx context.Context, name string, args ...string) (string, error)
    RunWithRetry(ctx context.Context, retries int, delay time.Duration, name string, args ...string) error
}

// DefaultRunner wraps exec.CommandContext with structured logging and error propagation.
type DefaultRunner struct {
    Logger *logrus.Entry
}

[...]

PR #267, opened 2026-07-19
security: limit all io.ReadAll calls to 1MB to prevent DoS

Amperstrand:
## Summary

6 of 7 `io.ReadAll` calls read HTTP response bodies without a size limit. A malicious upstream gateway or LNURL service could exhaust router memory by returning a multi-GB response.

The main HTTP handler (`main.go:413`) already limits request bodies to 1MB via `http.MaxBytesReader`. This PR applies the same limit to all response body reads.

## Changes

| File | Line | What's read | Limit |
|---|---|---|---|
| `lightning/lightning.go` | 71 | LNURL resolve response | 1MB |
| `lightning/lightning.go` | 115 | LNURL invoice response | 1MB |
| `upstream_session_manager/session.go` | 353 | Gateway payment response | 1MB |
| `upstream_session_manager/tollgate_prober.go` | 194 | Gateway probe response | 1MB |
| `upstream_session_manager/tollgate_prober.go` | 280 | Body drain (discard) | 1MB |
| `upstream_session_manager/upstream_usage_tracker.go` | 186 | Usage data response | 1MB |

All responses are small JSON (< 10KB expected). 1MB limit is generous.

## Why this matters

TollGate routers run on embedded hardware with 128-512MB RAM. An unbounded `io.ReadAll` on a malicious response could OOM-kill the process. The router connects to external services (LNURL providers, upstream TollGate gateways) that could be compromised or malicious.

## Verification

```bash
gofmt -l .     # clean
go build ./...  # OK
go test ./...   # all pass
```

Found during the codebase audit in #263.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ and [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- **HTTP response body reads now limited to 1 MB.** All `io.ReadAll`
+  calls on HTTP response bodies (LNURL resolve, invoice fetch, gateway
+  probes, usage tracker) now use `io.LimitReader` with a 1 MB cap,
+  matching the existing limit on the main payment handler. Prevents
+  OOM crashes on resource-constrained routers when a malicious or
+  compromised upstream service returns an oversized response
+  ([#267](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/267)).
+
 - **Lightning quote persistence across restarts.** Lightning invoice
   quotes are now persisted to disk (`quotes.json` in the wallet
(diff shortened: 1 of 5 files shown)


Amperstrand:
Reviewed the diff — this is correct and well-scoped. The `io.LimitReader(resp.Body, 1<<20)` wrapping is the right pattern, applied consistently to all 6 unguarded `io.ReadAll` call sites. The 1MB cap matches the existing `http.MaxBytesReader` on the payment handler. CHANGELOG entry is clean.

One note: if #268 merges first, there may be a minor CHANGELOG conflict (both add entries under `[Unreleased] → Fixed`). Easy rebase resolution.

Ship it.

PR #287, opened 2026-07-24
docs: correct NDS mark descriptions and fw4 priority diagnosis

Amperstrand:
Follow-up to #283 (merged). Two comment inaccuracies in the .nft file and CHANGELOG:

1. **`0x10000` is NDS "blocked" (admin-blocked client), not "pre-authenticated".** Pre-authenticated clients are unmarked (mark=0). Verified against nodogsplash source: [`src/conf.h:82`](https://github.com/nodogsplash/nodogsplash/blob/6beb760589da5fdeaa7b5f7c5277e38215811dbf/src/conf.h#L82): `DEFAULT_FW_MARK_BLOCKED 0x10000`.

2. **Diagnosis: nftables ordering at the same priority is undefined, not "inet accepts before ip filter".** Per the [netfilter mailing list](https://www.spinics.net/lists/netfilter/msg60396.html): "If exactly the same priority is used for two chains... order becomes undefined." The fix at priority -1 is correct — it establishes deterministic ordering.

Pure documentation change, zero functional impact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,13 +13,14 @@ and [Semantic Versioning](https://semver.org/).
 ### Fixed
 
 - **NDS fw4/nftables enforcement bridge.** NDS 5.0.2 inserts enforcement
-  chains in `ip filter FORWARD` (iptables-nft), but OpenWrt 24.10's fw4
-  `inet fw4 forward` chain accepts forwarded traffic at the same priority
-  before the ip filter chain runs — leaving NDS's enforcement chain dead
-  (0 packets). Authenticated clients had mangle marks set correctly but
-  traffic was never actually gated. Added `/etc/nftables.d/20-nds-enforce.nft`
-  include file that hooks into `inet fw4 forward` at priority -1 (before
-  fw4's accept) and enforces NDS marks. Bumped setup version to v0.5.1.
+  chains in `ip filter FORWARD` (iptables-nft) at priority 0. OpenWrt 24.10's
+  fw4 `inet fw4 forward` chain also runs at priority 0 — nftables does not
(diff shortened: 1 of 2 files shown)


Amperstrand:
## Verified: Mark descriptions need correction

Confirmed on our OpenWrt VM by examining actual nftables rules + NDS source:

### Current (incorrect) comments
The `.nft` file says `0x10000 = pre-authenticated`, but:

### Actual behavior (verified)
| Mark | NDS meaning | Matched by nft rule? |
|---|---|---|
| **0x0** (no mark) | Pre-authenticated (default state) | ❌ NOT matched — falls through to WAN reject |
| **0x10000** | Admin-blocked | ✅ Rule 1: `drop` |
| **0x20000** | Authenticated | ✅ Rule 2: `accept` |
| **0x30000** | Trusted | ✅ Rule 3: `accept` |

This means pre-authenticated clients (mark=0) are NOT caught by any of the three mark-based rules. They rely entirely on the catch-all WAN reject (rule 4) — which as #289 shows, is broken on non-GL-MT30 hardware due to hardcoded interface names.

**The PR is correct — the comments should be updated.** This is not just cosmetic: understanding the mark values correctly is essential for debugging why some clients bypass NDS enforcement.

Amperstrand:
## Fix in branch: Amperstrand:fix/nds-enforce-portable-wan

Mark descriptions corrected as proposed in this PR. The same branch also fixes #289 (portable WAN detection). PR creation pending due to GitHub API issues — will be created when API recovers.

Amperstrand:
Closing — superseded by #297 (merged). PR #297's branch corrected the same mark descriptions plus added portable WAN detection. The comment fixes from this PR are included in #297's diff.

PR #295, opened 2026-07-25
fix: portable WAN detection for NDS enforcement chain (#289)

### LINKED ISSUE #289: NDS enforcement chain: hardcoded WAN interface names break non-GL-MT3000 arches
## Problem

The `/etc/nftables.d/20-nds-enforce.nft` (PR #283) hardcodes WAN interface names:

```nft
meta nfproto ipv4 iifname "br-lan" oifname { "eth0", "phy0-sta0" } counter reject with icmp type port-unreachable
```

- `eth0` is GL-MT3000 WAN ✅
- `phy0-sta0` is GL-MT3000 wireless upstream (reseller mode) ✅
- Other arches (mips_24kc ath79 → WAN on `eth1`, x86_64 QEMU → single-interface `br-lan`, USB modems → `wwan0`/`usb0`, PPPoE → `pppoe-wan`) ❌

On affected arches, the reject rule for pre-authenticated traffic to WAN never fires. The drop rule for blocked clients (0x10000) still works regardless of oifname, but the captive portal redirect mechanism (reject → ICMP port-unreachable → OS CPD trigger) is dead.

## Confirmed on x86_64 QEMU

Tested on the local virtual lab: single-interface VM (br-lan only, no eth0). The reject rule counter showed 0 packets despite active pre-auth traffic.

## Suggested fix

Generate the oifname set from UCI in `99-tollgate-setup`:

```sh
WAN_DEV=$(uci get network.wan.device 2>/dev/null)
# Also add modem/wwan interfaces
for section in $(uci show network 2>/dev/null | awk -F'[.=]' '/^network\.modem_/ {print $2}' | sort -u); do
    MODEM_DEV=$(uci get network.$section.device 2>/dev/null)
    [ -n "$MODEM_DEV" ] && WAN_DEV="$WAN_DEV $MODEM_DEV"
done
# Generate .nft file with resolved interface names
```

Or match by destination scope instead of interface name:
```nft
meta nfproto ipv4 iifname "br-lan" ip daddr != { 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16 } counter reject with icmp type port-unreachable
```

## Priority

Follow-up — not blocking. GL-MT3000 (the primary target platform) works correctly. Other arches are affected but the drop rule for blocked clients still functions.

Amperstrand:
Replaces hardcoded `oifname { "eth0", "phy0-sta0" }` with `fib daddr oif != "br-lan"` — uses the kernel routing table instead of interface names.

**Before:** The reject rule for pre-auth traffic was GL-MT3000-specific. On mips_24kc (WAN on eth1), PPPoE (pppoe-wan), USB modems (wwan0), or any non-eth0 WAN, the rule was dead — pre-auth traffic to the internet was never rejected, so the OS captive portal detector never triggered.

**After:** The `fib daddr oif` lookup asks the kernel "what interface would this packet exit?" at evaluation time. If it's not br-lan (i.e., traffic is going to the internet), reject. Works on every architecture, every topology, every WAN type — now and in the future.

Verified: 12/12 PRTA NDS fw4 tests pass on OpenWrt 24.10.1 QEMU VM.

Closes #289.

### DIFF
--- a/packaging/files/etc/nftables.d/20-nds-enforce.nft
+++ b/packaging/files/etc/nftables.d/20-nds-enforce.nft
@@ -10,7 +10,7 @@
 #   0x10000 = pre-authenticated → drop
 #   0x20000 = trusted           → accept
 #   0x30000 = authenticated     → accept
-#   unmarked to WAN             → reject (force captive portal)
+#   unmarked (0) = pre-auth      → reject to WAN (force captive portal)
 #
 # Loaded by fw4 on boot and on `fw4 reload` / `/etc/init.d/firewall reload`.
 # Survives NDS restarts (independent of NDS process lifecycle).
@@ -28,5 +28,5 @@ chain nds_enforce_forward {
     meta nfproto ipv4 iifname "br-lan" meta mark & 0x00030000 == 0x00030000 counter accept
 
     # Unmarked traffic from br-lan to WAN — reject (captive portal redirect)
-    meta nfproto ipv4 iifname "br-lan" oifname { "eth0", "phy0-sta0" } counter reject with icmp type port-unreachable
+    meta nfproto ipv4 iifname "br-lan" fib daddr oif != "br-lan" counter reject with icmp type port-unreachable
 }


Amperstrand:
Closing as duplicate of #297 (same fix, cleaner commit message). Both change the same 1 file with the same oifname != br-lan fix.

Issue #88, opened 2026-04-20
Router-to-router autopay implicitly depends on nodogsplash client registration — breaks protocol compatibility


Origami74:
## Problem

When a downstream router pays an upstream TollGate on behalf of its clients (router-to-router autopay, [#77](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/77)), the upstream opens a gate for the paid MAC but then cannot measure usage, because that MAC has never generated LAN traffic on the upstream and is therefore unknown to `ndsctl`.

Concretely:
- `ndsctl json <mac>` returns `{}` for the paid MAC.
- [src/valve/valve.go:232-233](src/valve/valve.go#L232-L233) treats this as `"client with MAC X not found in ndsctl"`.
- [src/valve/customer_data_tracker.go:73-82](src/valve/customer_data_tracker.go#L73-L82) currently swallows the error and returns `(0, nil)` usage.

The net effect: the gate opens, usage is reported as 0 forever, the allotment is never reached, and the session never closes until the service is restarted. See [#78](https://github.com/OpenTollGate/tollgate-module-basic-go/issues/78) for a log reproduction.

## Why this is a protocol-compatibility bug

The TollGate payment protocol does **not** require that the paying MAC be pre-registered with the upstream's nodogsplash instance. A peer router (or any client speaking the protocol) should be able to pay for arbitrary downstream MACs without the upstream first having observed those MACs on its own LAN.

The current implementation has silently introduced an implicit dependency on nodogsplash's client table as a side effect of how \`ndsctl\` is used for bandwidth accounting. This:
- Breaks compatibility with peer TollGate implementations that don't happen to front nodogsplash.
- Makes router-to-router autopay effectively non-functional for bandwidth-metered sessions.
- Masks billing errors as \"0 usage\" rather than surfacing them.

## Options for a fix

1. **Proactively register the MAC with nodogsplash** when opening a gate — e.g. run \`ndsctl auth <mac>\` before the first stats query. Keeps the existing usage-tracking pipeline but removes the implicit \"must have been seen on LAN first\" constraint.
2. **Decouple usage tracking from nodogsplash** — use a MAC-agnostic source such as conntrack counters, nftables accounting, or an in-process packet counter. More invasive but drops the nodogsplash dependency entirely.
3. **Hybrid**: always authorize via \`ndsctl auth\` and fall back to alternative accounting if \`ndsctl json\` still returns empty after authorization.

## Acceptance criteria

[...]

PR #327, opened 2026-07-28
chore: add NUT spec quotes (greatspectations Layer 1)

Amperstrand:
## What

Adds verbatim NUT specification quotes as comments at key Cashu code locations, enabling CI-enforced spec drift detection via [greatspectations](https://github.com/rustyrussell/greatspectations).

## Quotes Added

| NUT | Quote Location | Spec Text |
|---|---|---|
| NUT-00 | `main.go` above `extractCashuToken` | Token format prefix: `cashu[version][token]` |
| NUT-00 | `tollwallet.go` above `Receive` | Verification: `k*hash_to_curve(x) == C` |
| NUT-05 | `tollwallet.go` above `Send` | Melt quote: `POST /v1/melt/quote/{method}` |

## Config

`specquotes.toml` configured for greatspectations:
```toml
[sources.NUT]
format = "markdown"
dir = "nuts"
pattern = "{id:02d}.md"
```

## Context

Part of the [cashu-audit](https://github.com/Amperstrand/cashu-audit) framework — Layer 1 adoption. The gonuts-tollgate library already has 85 spec quotes. This adds TollGate's own Cashu handling code to the same framework.

## Verification

- `go build` passes
- `gofmt` clean
- Comments only, no code changes

PR #340, opened 2026-07-30
feat: token recovery tool + AI audit prompts + speccheck

Amperstrand:
## Token Recovery Tool

Standalone Go tool (`scripts/token-recovery/`) that recovers Cashu tokens from the `tokens-to-recover.txt` file written when autopay fails (e.g., NDS #88 gate-open bug).

### How it works
1. Parses each line: `timestamp | mint_url | token | error`
2. Decodes token (V3/V4)
3. Checks proof state at mint via `POST /v1/checkstate` (NUT-07)
4. If UNSPENT → calls `Wallet.Receive(token, swapToTrusted=true)` 
5. If SPENT → skips (value already consumed)
6. Reports: X recovered, Y spent, Z failed

### Usage
```bash
# Dry run — check what is recoverable
go run scripts/token-recovery/ -file /etc/tollgate/tokens-to-recover.txt -dry-run

# Actually recover
go run scripts/token-recovery/ -file /etc/tollgate/tokens-to-recover.txt

# Cross-compile for router
CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -o recover scripts/token-recovery/
scp recover root@router:/tmp/
ssh root@router /tmp/recover


Amperstrand:
Cold-review item 3 + housekeeping: **resolved** (`8ef6a26`, CI green, PR undrafted).

**speccheck exit semantics — fixed here, with a corrected diagnosis.** The script had no `exit $?` (that part of the finding doesn't match the file); the actual mechanism is `set -e` + a non-zero exit from `spectate check` when drift is found, which aborted before the coverage report and propagated as the script's status. Now: `|| true` on both report commands + explicit `exit 0`; infra failures (install/clone) still abort via `set -e`. Functionally verified with a stub `spectate` exiting 1 on drift: script reports and exits 0.

**Consolidation:** this PR is now the single home of the audit tooling — `make/speccheck.sh` + `docs/ai-audit/` were previously duplicated (and diverging: `greatspectate` vs `spectate` command names) across this branch and #327; #327's rebuild dropped its copies. Note also this PR's tree never contained the ELF binary (`7de4e476`) — that was #341/#343 only, both since rebuilt clean.

Amperstrand:
Cold-review response — both majors + minors addressed in `03ccf14` (CI green):

- **Major 1 (PENDING treated as recoverable): confirmed real, fixed.** Recoverable now requires `State == Unspent` for every proof; PENDING gets its own status (⏳, retry-later message) — receiving in-flight proofs was a double-spend attempt.
- **Major 2 (wallet per line): fixed** — one lazily-loaded wallet handle for the whole run (`LoadWallet` locks the bolt store; per-line loads contended). `Receive` operates on each token's own mint; first recoverable token's mint sets the swap-to-trusted default.
- Minors: dry-run recoverables now counted as 💰 not ❌; `p.Secret[:16]` removed from error output entirely (secrets spend tokens — also fixes the short-secret panic); **no timeout change needed** — the gonuts client already wraps all HTTP in a 30s `http.Client` timeout (`wallet/client/client.go:34`), so the finding was stale. README summary line updated for the new statuses.

PR #341, opened 2026-07-30
chore: gofmt cleanup — 16 files

Amperstrand:
Pure formatting. `gofmt -l .` now returns empty on all source files.

### DIFF
--- a/src/cli/merchant_provider_test.go
+++ b/src/cli/merchant_provider_test.go
@@ -22,14 +22,14 @@ func (m *namedMerchant) DrainMint(mintURL string) (string, uint64, error) {
 	return "", 0, nil
 }
 func (m *namedMerchant) GetAcceptedMints() []config_manager.MintConfig { return nil }
-func (m *namedMerchant) GetBalance() uint64                             { return 0 }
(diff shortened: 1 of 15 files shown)


Amperstrand:
Cold-review item (committed ELF binary): **resolved.** The 13MB unstripped `scripts/token-recovery/token-recovery` blob (`7de4e476`) had entered via an unrelated commit on this branch line, not the gofmt work itself. Branch rebuilt onto current main carrying only the gofmt cleanup (`1e06f22`) — binary absent from the diff, CI green. Note for the record: the blob was never in #340's tree; only this branch and #343's carried it.

Amperstrand:
Cold-review response — the import-ordering fix landed as `7b124b0` (yours — thank you, faster than my queued patch; local synced to it). On the goimports-vs-gofmt consistency minor: the remaining two files' changes are goimports-style import grouping, which the current hook doesn't enforce. Options: (a) leave as-is — non-blocking cosmetic; (b) extend #343's hook with a goimports check (needs goimports on contributor machines / CI). Suggest (a) for now with a follow-up issue if you want the repo goimports-clean; say the word and I'll file it.

PR #336, opened 2026-07-28
docs: fix stale SSID format comment

Amperstrand:
One-line comment fix. References #229 (closed — format already implemented in code).

PR #343, opened 2026-07-30
chore: add gofmt check to pre-commit hook

Amperstrand:
Extends `hooks/pre-commit` to also check `gofmt -l` on staged `.go` files. Prevents the 16-file gofmt cleanup situation from recurring.

Combined with the existing dependency sync check, the pre-commit hook now catches:
1. Dependency version drift across go.mod files
2. Unformatted Go source files

Install: `git config core.hooksPath hooks`

Also recommend adding gofmt to CI (test.yml):
```yaml
  gofmt-check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v6
      - uses: actions/setup-go@v6
        with:
          go-version-file: src/go.mod
      - name: gofmt check
        run: |
          cd src
          if [ -n "$(gofmt -l .)" ]; then
            echo "::error::gofmt found unformatted files"
            gofmt -l .
            exit 1
          fi
```

### DIFF
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -13,9 +13,10 @@ if git diff --cached --name-only | grep -q '\.go$'; then
     python3 tests/contract/check-import-paths.py
 fi
 
-# Check gofmt on staged Go files
+# Check gofmt on staged Go files (--diff-filter=d: gofmt cannot read
+# deleted files; xargs -I{} keeps one file per line, spaces-safe)
 if git diff --cached --name-only | grep -q '\.go$'; then
-    unformatted=$(gofmt -l $(git diff --cached --name-only -- '*.go'))
+    unformatted=$(git diff --cached --name-only --diff-filter=d -- '*.go' | xargs -I{} gofmt -l -- {})
     if [ -n "$unformatted" ]; then
         echo "❌ gofmt found unformatted files:"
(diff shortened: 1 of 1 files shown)


Amperstrand:
Cold-review item (committed ELF binary): **resolved.** Same blob as #341 (`7de4e476`, 13MB unstripped) — entered via an unrelated commit under the hook work. Branch rebuilt onto current main with only the gofmt-hook commit (`a911079`); binary gone, `bash -n` + functional gofmt-check verification re-run (flags a deliberately unformatted staged file, passes a clean one), CI green. Side effect of the rebuild: no longer stacked on #336's SSID commit — the diff is now the hook alone, so #341→#343 ordering pressure is gone.

Amperstrand:
Cold-review response — major + minor fixed in `1eef16d` (CI green):

- **Major (staged deletion crashes hook): confirmed by inspection, fixed.** `--diff-filter=d` excludes deletions from the gofmt list; files are fed one-per-line via `xargs -I{}` (spaces-safe, and portable — no `mapfile`, so it works in OpenWrt's ash too). Functionally verified: staged deletion + unformatted file flags exactly the unformatted file, exit 1, no crash.

c03rad0r (approved):

PR #360, opened 2026-08-26
fix(merchant): contain payment-goroutine panics

Amperstrand:
## Context

`PurchaseSession` runs the wallet `Receive` call in a goroutine with no `recover`:

```go
go func() {
    amount, err := m.tollwallet.Receive(paymentCashuToken)
    ch <- receiveResult{amount, err}
}()
```

Any panic in the wallet layer (e.g. a mint serving malformed keysets — see the empty-proofs `proofsToSwap[0]` class of bugs) therefore **kills the entire `tollgate-wrt` process**, taking down every concurrently active customer session, and can be triggered by any unauthenticated portal client that submits a crafted token.

## Fix (W-1b: panic containment)

The goroutine now defers a recover that **sends an explicit error into the existing buffered channel**:

```go
defer func() {
    if r := recover(); r != nil {
        ch <- receiveResult{0, fmt.Errorf("payment processing panicked: %v", r)}
    }
}()
```

Fast-error contract: the caller's `select` (which keeps its 30s timeout arm unchanged) receives this result immediately and returns a signed `payment-processing-failed` notice — instead of process death or a misleading 30-second `payment-processing-timeout` notice.

Channel safety: a panic can only fire before the normal send, so the deferred send never double-sends; `ch` is buffered(1), so the deferred send never blocks even if the caller has already taken the timeout arm.

## Test

`TestPurchaseSessionPanicContainment` (src/merchant/purchasesession_panic_test.go) drives the real `PurchaseSession` with a stub `WalletPort` whose `Receive` panics, and asserts:

- a notice event of kind 21023 with code `payment-processing-failed` comes back,
- its content contains `panicked`,
- it all completes in **< 5 s** (deadline enforced inside the test — proves the fast-error path, not the 30s timeout arm),
- the test process survives (panic contained, not propagated).

On unpatched code this test crashes the test binary (`panic: wallet exploded ... at merchant.go:475` in `PurchaseSession.func1`) — i.e. it demonstrably bites.

Note: running tests inside `src/merchant` requires `go mod tidy` first — a pre-existing go.mod staleness on `main` (gonuts pinned at v0.10.0 while the code imports v0.11.1), unchanged by this PR and out of scope here.


Amperstrand:
⚠️ **Branch cleaned up (force-push) — please review the new head.**

This branch (and #361, #362, #363) was based on pre-incident history and still carried `deploy-backup-20260730/` — the router backup with the **merchant private identity key, wallet.db and recovery tokens** that were purged from main via the history rewrite (see the incident report at main@HEAD and #364). Merging as-is would have re-committed the leaked key.

The branch has been rebased onto rewritten main (`6c974f7`) and now carries **only the panic-containment change**:

- `src/merchant/merchant.go` — `defer recover()` in the payment goroutine sends the error into the existing buffered channel
- `src/merchant/purchasesession_panic_test.go` — unchanged
- CHANGELOG entry

**Dropped:** the secrets directory, the `DEPENDS` packaging change (`nodogsplash,luci,jq` → `libc`), `.pre-commit-config.yaml`, `packaging/local-build-ipk.sh`, the token-recovery binary, and unrelated go.mod churn. If any of those are wanted, please propose them as separate PRs — the DEPENDS change in particular needs its own discussion (it removes the declared nodogsplash/luci/jq deps from the ipk).

Verified locally: build clean, and `TestPurchaseSessionPanicContainment` passes on the cleaned branch (panic contained, `payment-processing-failed` notice in <5s). E2E cloud-lab run on this head is in flight; results will follow.

c03rad0r (approved):

Amperstrand:
**E2E results (follow-up to review):** deployed head `bbfd84d` as `ci-pr-360.197.bbfd84d` (CI-built ipk) on the OpenWrt x86_64 QEMU lab:

| Suite | Result |
|---|---|
| api/test_quote_persistence | **4/4 passed** |
| api/test_lightning_backoff | **3/3 passed** |

The panic-containment change itself carries the strongest evidence in its unit test — re-verified on the cleaned branch: `TestPurchaseSessionPanicContainment` passes (panic contained, `payment-processing-failed` notice returned in <5s, test process survives). Combined with the green suites above on the built binary, this is merge-ready from our side.

PR #383, opened 2026-09-13
feat(build): reproducible package builds — pinned inputs, SOURCE_DATE_EPOCH, reproducibility test

Amperstrand:
## What this does

Makes `tollgate-wrt` artifacts **byte-for-byte reproducible**: given the same source revisions, pinned toolchain versions, target arch, format, and compression, two clean builds produce identical SHA-256s. Also lands three supporting fixes that surfaced on the way (`--version` on both binaries, license metadata, `.gitignore` anchoring).

Verified on a 20-core build host, each artifact built in **two independent clean roots** (fresh tree copies, isolated `HOME`/Go/npm caches — cache state cannot make hashes match):

| Artifact | SHA-256 (build 1 = build 2) |
|---|---|
| `tollgate-wrt` / `tollgate` (x86_64) | `0b825ec4…` / `ca694f02…` |
| portal tree (pinned SHA `86ac5fc7…`) | `88c6f84f…` |
| ipk x86_64 / aarch64 | `7fa7a658…` / `14b95309…` |
| ipk x86_64 + UPX ultra-brute (5.2.1) | `7b1220a1…` |
| apk x86_64 (SDK digest-pinned) | `13b39c18…` — no upstream blocker |

## How (17 commits, grouped)

- **Pinned inputs**: `packaging/build-inputs.json` (Go 1.25.8, Node 22.17.0, npm 10.9.2, UPX 5.2.1 — each with verified tarball sha256; portal commit; six OpenWrt SDK registry digests) + `packaging/build-env.sh` as the single loader (`SOURCE_DATE_EPOCH` default = HEAD commit timestamp, deterministic `BuildTime`, centralized ldflags, strict-mode fail-closed).
- **Deterministic packers**: `build-ipk.sh` (sorted members, epoch mtimes, `gzip -n` — gzip header timestamps were a real leak), `local-build-ipk.sh` (per-arch, UPX-aware, toolchain checks), `portal-build.sh` (floating `main` ref → pinned SHA; Node 20 CI mismatch → portal's declared 22.17.0), `build-sdk-package.sh` (digest-pinned image, epoch + staged-mtime normalization; also adds the missing x86-64 target).
- **Test + CI**: `scripts/repro-test.sh` / `make reproducibility-test T=<binaries|portal|ipk|ipk-upx|apk> ARCH=<arch>` (two clean roots, sha256, diffoscope on mismatch); `repro-check.yml` (fast binary lane on push/PR, package lane on dispatch); `build-package.yml` now derives `BuildTime` from the source epoch, pins Go/Node/UPX/SDK digests, adds `-buildvcs=false`.
- **Fixes en route**: `--version` flag on both binaries (cobra `Version` field + daemon early-exit above the fatal config init; also fixes the CLI binary having *no* embedded version — its separate Go module never linked the `src/cli` ldflags target), license metadata `CC0-1.0` → `GPL-3.0-only` (matches `LICENSE`), `.gitignore` binary patterns anchored (bare `tollgate-cli` silently ignored new files in `src/cmd/tollgate-cli/`), hosted-runner-independent `scripts/shc-build-package.sh` (VM order→build→verify→reap; used while Actions was down).
- Docs: `docs/reproducible-builds.md` (inputs table, epoch choice, how to test, how to bump pins, how to reproduce a release).

## Things a reviewer would ask

- **New build-time dep: `jq`** — needed by `build-env.sh` to read the manifest. Host-side only; no router-runtime change.
- **gofmt stragglers commit** touches 3 files unrelated to the feature because the pre-PR gate (`gofmt -l .` must print nothing) currently fails on `main`; not a drive-by.
- **Overlap with #380** (single-source the version): both touch `build-package.yml`, `local-build-ipk.sh`, `build-sdk-package.sh`, and version stamping. This PR centralizes *epoch/ldflags*; #380 centralizes the *version string*. Whichever merges first, the other rebases on top — happy to rebase onto #380 if it lands first.
- **Known limits** (in docs): reproducibility verified same-host; cross-host (GH runner vs build host) should hold because ordering/mtimes/owner/gzip headers are normalized explicitly, but is not yet exercised (org Actions suspension). Coverage is x86_64-all-formats + aarch64 ipk; remaining arch/UPX variants use the same mechanism.

## Test plan

- [x] `gofmt -l .` clean; `go vet`/`go build`/`go test -race -count=1 -tags testenv ./...` green (root + CLI modules)
- [x] `tests/contract/build-purity.sh` 3/3; `js-schema-lint` PASS
- [x] `make reproducibility-test` × {binaries, portal, ipk×2 arches, ipk-upx, apk} → `REPRODUCIBLE: YES` (hashes above)


Amperstrand:
**Post-rebase re-verification** (branch tip `c721aa8`, after rebasing onto `d6e4e09` and adding the changelog links): the full fast battery was re-run in two fresh clean roots on the build host — binaries, x86_64 ipk, and the x86_64 apk all report `REPRODUCIBLE: YES`.

Note the absolute hashes differ from the table in the PR body (`79a95aaf…`/`4154a009…` for the binaries, `2a699c26…` for the ipk, `387abd32…` for the apk). That is the system working as designed, not drift: the branch tip is a new commit with a new timestamp, so the `SOURCE_DATE_EPOCH`-derived `BuildTime` legitimately changed. Identity is per-commit; within every run the two independent roots agree exactly.

Standing offer from the body still applies: when #380 lands, this branch rebases onto it and switches the default package version to read the new `VERSION` file, then re-runs the same battery on the new epoch.

Amperstrand:
**External alignment review** — checked this PR against the reproducible-builds.org spec, OpenWrt's own build system, and industry container practice (research notes in the thread below).

## Where this matches the standards exactly

- **`SOURCE_DATE_EPOCH` default**: the [spec](https://reproducible-builds.org/specs/source-date-epoch/) recommends *"the date of the latest commit in their version control system"*, and OpenWrt's own `scripts/get_source_date_epoch.sh` does byte-for-byte the same thing we do — `git log -1 --format=%ct`. Our `build-env.sh` default is the canonical derivation, not an invention.
- **Spec conformance details**: malformed-epoch values make us exit non-zero (spec: SHOULD); the variable is exported, never unset, for children (spec: MUST); ASCII-integer validation matches. 
- **`-buildvcs=false -trimpath`**: identical to `lang/golang/golang-package.mk`'s `GO_PKG_INSTALL_ARGS` in openwrt/packages — upstream's own Go packaging flags. When tollgate moves to a source-built package Makefile, the flags already agree.
- **Package mtime normalization**: OpenWrt's reproducible-builds work uses exactly our technique (*"Created packages are `touch`ed with a specific timestamp so resulting packages have the same checksums"* — commit a83b2af) and their `prepare_rootfs` normalizes all file timestamps when `SOURCE_DATE_EPOCH` is set. Our staged-tree `touch` + `--mtime=@$epoch` is the same play.
- **Digest-pinned SDK images**: `name:tag@sha256:…` is the documented immutability pattern (Docker/Chainguard); we are stricter than OpenWrt's own docker examples, which float on tags — appropriate for byte-reproducibility. Notably, the index-vs-config-digest trap we hit and fixed during development is the exact failure mode Chainguard's tooling docs warn about; fetching via the registry API was the correct resolution.
- **`npm ci` against the pinned portal lockfile** matches standard reproducible-container guidance (lockfile + `npm ci`, never `npm install`).

## Two findings from upstream's current direction (worth knowing, neither blocking)

1. **OpenWrt is debating moving package `SOURCE_DATE_EPOCH` to a fixed value** ([openwrt#21579](https://github.com/openwrt/openwrt/issues/21579) → PR #21587, "set SOURCE_DATE_EPOCH to 0", Jan 2026) because per-package git derivation breaks in shallow feed clones and the SDK. Two consequences for us: (a) it validates this PR's design that `build-env.sh` *requires* an explicit epoch when git history is absent — the exact upstream failure mode is impossible here; (b) when tollgate is eventually built inside openwrt/packages, their epoch policy (possibly 0) will govern buildbot artifacts, so our feed artifacts may intentionally differ from theirs by epoch. That's per-channel identity, not nondeterminism — but worth a line in the docs when upstreaming.
2. **We set uniform mtimes rather than clamping.** The spec offers `--clamp-mtime` (preserve older timestamps, clamp newer); upstream's direction (uniform fixed timestamp) also abandons preservation. For a package whose payload is freshly built binaries + staged files, uniform-at-epoch is the simpler equivalent and matches where OpenWrt is heading. Noted as a conscious choice.

## Minor follow-ups surfaced by the research (not in this PR)

- Digest/tool refresh is manual today; the industry pattern is a small update script over the manifest (like a lockfile bot). Cheap to add later.
- The spec's "defer formatting to runtime for user-observable values" is deliberately not applied to `BuildTime` — embedded UTC build timestamps in binaries are the universal exception (Go ecosystem-wide); our `version` output is UTC by convention.

**Disposition: unchanged — land.** The mechanisms are the same ones OpenWrt itself uses, the deviations are conscious and documented, and the two upstream-movement findings are forward-compat notes rather than defects.

c03rad0r (approved):

Amperstrand:
**Re-verification after the #380/#379 merges and the #386 stack** (branch tip `c4a3938`): full battery re-run in two fresh clean roots on the build host, now with the version flowing from the new `VERSION` file (`v0.6.0-alpha2`):

| Artifact | BUILD 1 = BUILD 2 |
|---|---|
| `tollgate-wrt` / `tollgate` (x86_64) | identical — `REPRODUCIBLE: YES` |
| ipk x86_64 (`…v0.6.0-alpha2_x86_64.ipk`) | `47075124…` — `REPRODUCIBLE: YES` |
| portal tree | `REPRODUCIBLE: YES` |
| apk x86_64 (`0.6.0_alpha2-r0.apk`) | `bcc0b4c9…` — `REPRODUCIBLE: YES` |

Hashes changed from prior runs for the expected reasons: new epoch (branch tip), new version string (VERSION single-sourcing from #380), plus the deterministic improvements stacked by #386 (locale/TZ pinning, mode-bit normalization, Section/Source control fields). Within every pair the independent roots agree byte-for-byte — the stacked hardening did not disturb determinism, and the VERSION-based versioning path is verified end-to-end including the apk pre-release normalization (`0.6.0_alpha2-r0`).

Amperstrand:
@felixfelix-bot — re: your three follow-ups, all three have landed, plus your original blocker-1 suggestion, plus the hash-equality evidence your item 2 implied. Requesting a re-review at `48a313f`.

**1. Zero-comparison guard** — landed via #386: `scripts/repro-test.sh` now counts artifact pairs actually compared (`comparisons` counter) and gates `REPRODUCIBLE: YES` on it; a root that yields no artifact (or an empty portal tree) sets `fail=1` via `no_artifacts()` instead of passing vacuously. Every recent run prints `compared N artifact pair(s)`.

**2. Harness builds what release builds** — landed via #386's unification: the release compile step now sources `packaging/build-env.sh` and stamps `VERSION` through the same `go_ldflags`/`cli_ldflags` helpers, and the harness derives `TG_GIT_COMMIT` from the real repository HEAD (never a placeholder). Evidence that the two paths now converge — built via the exact release-recipe commands and via the harness on the same checkout, and compared byte-for-byte on the build host:

- `tollgate-wrt` — release recipe `ab9fb1c2…` = harness `ab9fb1c2…` (**byte-identical**, `cmp` clean)
- `tollgate` CLI — identical likewise (`ee792c4a…` both)

One honest anomaly from the verification: with `HOME=/root` on that specific build box, the same commands produce a different (still internally deterministic) binary — `091ef9bd…` vs `ab9fb1c2…` — with fresh caches and identical env otherwise. The equality above holds under the harness's pinned env profile (isolated `HOME`, explicit `GOCACHE`/`GOMODCACHE`, strict inputs), which is exactly what CI runners have. I could not pin the perturbing variable inside the `/root` profile to a named `go env` setting; flagging it rather than hand-waving it.

**3. The lane executes** — `.ngit/act/workflows/repro-check.yml` exists alongside the GitHub workflow, so the fast binary reproducibility check runs on the CI that actually works.

**Plus your blocker-1 suggestion, now implemented** (`48a313f`): `define-package-matrix` no longer carries hand-copied digest literals — it checks out the repo and injects every row's SDK digest from `packaging/build-inputs.json` at generation time (fail-closed if a target is unpinned), so the matrix and the manifest structurally cannot drift. The digest literals are gone from the workflow entirely.

Full battery at the previous tip (`c4a3938`, content unchanged by `48a313f` — workflow/docs only): binaries, portal, ipk x86_64 (`47075124…`), and apk x86_64 (`bcc0b4c9…`) all `REPRODUCIBLE: YES` in two fresh clean roots, with the version flowing from `VERSION` (`v0.6.0-alpha2`, apk-normalized `0.6.0_alpha2-r0`).

Amperstrand:
**Real-OpenWrt verification of the reproducible artifact** (via the org's `physical-router-test-automation` framework, QEMU venue on KVM: OpenWrt **24.10.1** x86_64, the ipk lane's OS).

The x86_64 `.ipk` from the two-clean-roots reproducibility run was transferred to the router and verified end-to-end:

```
router# sha256sum /tmp/tollgate.ipk
47075124ddd8ef024726e164e3930227f8e38f0e7927141fd78a7b73b9a7466f   # == both build roots on the build host

router# opkg install /tmp/tollgate.ipk
Configuring tollgate-wrt.
Post-installation script completed successfully

router# opkg list-installed | grep tollgate
tollgate-wrt - v0.6.0-alpha2

router# tollgate --version && tollgate-wrt --version
tollgate version v0.6.0-alpha2
tollgate-wrt v0.6.0-alpha2

router# ls /etc/nftables.d/
10-custom-filter-chains.nft  20-nds-enforce.nft  30-backend-firewall.nft  README
```

So on a real OpenWrt system: the artifact's bytes match the reproducible build exactly (third machine, post-transfer), `opkg` installs it cleanly with the postinst completing, both binaries report the `VERSION`-file version, the daemon starts under procd, and the complete nftables ruleset ships (`20-nds-enforce` + `30-backend-firewall` via the local path — #387 fixes the same set for the CI lanes).

Also landed on this branch (`8a49cc3`): `scripts/update-build-inputs.sh` — a drift guard that verifies every pinned SDK digest and Go/Node tarball hash against the live registries (currently zero drift), with an `update` mode for the mechanical values; plus the upstream `SOURCE_DATE_EPOCH`-policy note in the docs (openwrt#21579 — feed vs buildbot hashes will intentionally differ).

PR #24, opened 2025-07-02
@Origami74's upstream-purchase implementation plan

c03rad0r:
I love the plan, just making sure there is a pull request so we can find the active branches easily :)

I'll try to demo a simple crows nest [here](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/11), but its just a demo, we can refactor it any time...

PR #23, opened 2025-07-01
check all fields of install config

c03rad0r:

PR #392, opened 2026-09-13
fix(wallet): bump gonuts to the #23/#24/#25 integration ref; make ErrTokenAlreadySpent reachable

Amperstrand:
Stacked on #383 (the release cuts from that branch; this PR is the intended next commit before the tag).

## What

Pins `gonuts-tollgate` to `717c886` (branch `tmp/release-integration` on the gonuts repo = main + #23 + #25 + #24) across every consuming module (root, merchant, cli, tollwallet), to be re-pinned to the tagged release the moment it exists. Riding in:

- **Empty-proofs tokens fail normally** instead of a contained panic (gonuts #23) — reachable today from the portal payment path; #360's recovery contains it, but a plain error is correct.
- **Mint swap rejections surface verbatim** instead of empty `could not swap proofs:` errors (gonuts #25) — and that is what finally makes `ErrTokenAlreadySpent` reachable, because:
- **this PR broadens the sentinel match.** tollwallet matched only the exact string `Token already spent`; mints say `inputs have already been spent`. Now case-insensitive `already spent` / `already been spent`, pinned by a phrasing table **and** a two-layer test that drives a mock mint's rejection through gonuts into `errors.Is(err, ErrTokenAlreadySpent)` — failing if either layer regresses to swallowing.
- Below-fee tokens fail fast; second in-process wallet load errors clearly instead of deadlocking (gonuts #24 — latent for tollgate, harmless).

## Verification

- gofmt / vet / build clean; root module `-race -tags testenv` green (15.3s); merchant (185s), cli, tollwallet modules green.
- E2E on the build host: cloud-lab (cdk-mintd FakeWallet mint, smoke payments) + the reproducibility battery at this head — results posted below before this is mergeable.


Amperstrand:
Review pass against PR-REVIEW.md — focusing on the **actual new content** (the tollwallet sentinel work), since ~2000 lines of this diff are #383's packaging/build content that is already on `main` via the #383 merge and will disappear on rebase.

**What this PR actually changes** (after the #383 noise is rebased away):

1. **`gonuts-tollgate` dependency pinned to `717c886`** across all modules — the integration ref that fixes empty-proof panics (#23), surfaces mint-swap rejections verbatim instead of swallowing them (#25), and fixes a second-load deadlock (#24). The description is candid that this is a pre-tag pin to be re-pinned to a release.

2. **`src/tollwallet/tollwallet.go` — sentinel broadening**: the old exact-match `strings.Contains(err.Error(), "Token already spent")` only caught Nutshell-family mints; CDK-family mints say "inputs have already been spent". The new `isAlreadySpentError()` helper matches both phrasings case-insensitively, and — critically — the match was **unreachable before the gonuts bump** because older gonuts swallowed the rejection into an empty error. This is a real bug fix: the `ErrTokenAlreadySpent` sentinel existed but could never fire for CDK mints.

3. **Test coverage is excellent** — two layers, each catching a different regression:
   - `TestIsAlreadySpentError_MintPhrasings` (unit): phrasing table with Nutshell, CDK, case variations, plus the guards — empty swallowed errors must NOT match (or every failed swap becomes "already spent"), unrelated errors must not match, nil is handled.
   - `TestReceive_SpentTokenMapsToSentinel` (integration): mock mint serves CDK-style `{"code":3,"detail":"inputs have already been spent"}` on `/v1/swap`; the test drives the rejection through gonuts into `errors.Is(err, ErrTokenAlreadySpent)`. Fails if either layer regresses (gonuts swallowing again, or match drift here).

**Blocking item (mechanical, not content):**

**Rebase onto `main`.** This PR was stacked on `feat/reproducible-builds` (#383), which merged as `f2a4728`. The diff against `main` currently shows ~2000 lines of #383 content (workflow pins, SDK digests, repro-test, portal pinning, `--version` flag, license fix) as if it were new. After rebasing onto `main`:
```bash
git rebase --onto main feat/reproducible-builds bump/gonuts-integration
```
the diff will show only the actual changes: the go.mod/sum bumps + the tollwallet sentinel work. Reviewers should not have to mentally subtract 35 files of noise.

**Non-blocking notes:**

- The broad `"already spent"` substring match is intentionally permissive — any mint error containing that phrase maps to `ErrTokenAlreadySpent`. The empty-error guard is the right defense, and the phrasing table pins the boundary. Worth a one-line comment in a future PR if mints appear with false-positive phrasings.
- The pre-tag pin (`tmp/release-integration` branch) is acknowledged in the description and carries a re-pin commitment. When the gonuts tag lands, that's a mechanical `go.mod` bump.

**Disposition: land-after-rebase.** The sentinel fix is real and the two-layer test design is exactly right — it catches regression at either the dependency layer or the integration layer, and the empty-guard prevents the broadened match from over-firing.

c03rad0r (approved):

PR #405, opened 2026-09-17
build(packaging): pin tar format to gnu in build-ipk.sh

Amperstrand:
Follow-up to the equivalence spike on #371 (SDK oracle vs SDK-free lanes).

**Change:** add `--format=gnu` to the three `tar` invocations in `packaging/build-ipk.sh`, matching buildroot's `scripts/ipkg-build` which passes it explicitly.

**Why:** the SDK-free ipk lane currently relies on GNU tar's *compile-time default* format. On any host whose tar defaults to pax/posix the output would silently differ — still opkg-readable, but not byte-reproducible vs the SDK lane. This was the last unpinned determinism knob on the ipk container path.

**Verification:** byte-neutral on the equivalence build host (ai-legion) — identical ipk sha256 (`dd3b0d37…`) built with and without the flag, same pinned inputs (aarch64_cortex-a53 @ `VERSION` v0.6.0-alpha2, main @ 67d4f6c).

**Deliberately out of scope (post-release, see #371):** consolidating the arch→(GOARCH/GOARM/GOMIPS/SDK) mapping that lives in build-package.yml / local-build-ipk.sh / repro-test.sh, and the license-text staging divergence (Amperstrand fork #92 — team decision).

cc @felixfelix-bot — please review; team flow is merge on your approval.

Issue #371, opened 2026-08-28
Audit: is the full OpenWrt SDK needed for .apk package releases, or only for OS image releases?


Amperstrand:
## Context (from the 2026-08-27 billing/minutes audit)

This repo is the org's largest Actions consumer (43,120 runner-minutes YTD). Two packaging paths exist with wildly different costs:

| Path | Design | Cost per leg |
|---|---|---|
| `.ipk` (6 targets × compression variants) | **SDK-free**: static Go binaries from Blossom + `packaging/build-ipk.sh` (plain tar/ar), `DEPENDS=libc` declared as metadata | minutes — cheap |
| `.apk` (2 targets) | **Full `openwrt/sdk:<sdk>-25.12.0` container**, `make package/tollgate-wrt/compile` with `CONFIG_PACKAGE_nodogsplash=y luci=y jq=y USE_APK=y` — i.e. rebuilds the LuCI/nodogsplash dependency chain from feeds per leg per run | tens of minutes — the sink |

Interim mitigation shipped in #370 (SDK dl/staging/build caching + timeouts). This issue is the follow-up audit.

## Question 1 — why does the apk path build dependencies at all?

Hypotheses to verify:
1. **Metadata validation only**: on a real router, `apk add tollgate-wrt` resolves dependencies from *official* OpenWrt feeds at install time. If true, building them in-CI validates our Makefile's DEPENDS against reality — useful, but perhaps only needed on releases, not every push.
2. **Real .apk dependency artifacts required**: if we ship/store the produced dep .apks (e.g. for offline/OS-image assembly), full builds are load-bearing and cannot be dropped.
3. **Historical**: the apk path pre-dates the SDK-free ipk path and inherited the heavyweight default.

Evidence to collect: does anything consume `/builder/bin/packages/*` beyond our own tollgate-wrt .apk? (The workflow only collects `find /builder/bin/packages -name '*.apk' | head -n1` — suggests only ours.)

## Question 2 — split by artifact class (maintainer's framing)

- **OS/image releases**: likely DO need full SDK (image assembly, real dep graphs).
- **Package releases (.ipk/.apk of tollgate-wrt only)**: plausibly SDK-free — same pattern as the .ipk path and as jmcorgan/fips ("one small C tool, no OpenWrt SDK"): static musl binaries + plain apk packaging (tar + signing).

## Proposed feature flag

`workflow_dispatch` + repo variable `SDK_MODE`:

| Value | Behavior |
[...]

PR #406, opened 2026-09-17
ci(publish): verify channel publication + require 3 mirrors on tags

Amperstrand:
Makes the release channel publish self-verifying. Companion to #405 and the equivalence evidence on #371.

**What it does**
1. New `verify-publication` job after `publish-metadata`: every (arch, format) the build matrix produced must have a kind-1063 event on the relays for the published version+channel; artifacts must be servable from ≥2 mirrors with the sha256 from the `x` tag (sample mode: one ipk + one apk, `VERIFY_DOWNLOAD=all` for full). `trigger-build-os` now depends on it — OS builds fire only on verified publications.
2. `BLOSSOM_MIN_SUCCESS` raised to 3 for tag refs (stays 1 for branch/dev builds).

**Why now:** both silent failure modes have already happened — `v0.6.0-alpha1` published nowhere (tag during the Actions outage: zero events, zero assets) and `v0.5.0` mirror rot (2 of 3 mirrors 404 today). With the final tag approaching (#339), this converts them into red builds instead of post-hoc archaeology.

**Functional evidence (run against the live channel):**
- v0.5.0/stable, strict ≥2 → FAIL with exactly the real state (`primal.net` sha256-ok, `psbt.me`/`cashu.email` down) — it would have caught the v0.5.0 incident at publish time
- v0.5.0/stable, lenient ≥1 → PASS
- matrix artifact missing from channel → FAIL (`expected artifact missing`)
- alpha1 (published nowhere) → FAIL (`zero events match`)

**Design notes:** version filtered client-side, channel relay-side — combined relay-side v/A filters are unreliable on some relays, and the relay-side channel filter bounds the newest-N window so high-frequency dev builds cannot crowd out older stable events (found live: dev events flooded the window). Runner cost: one small job, ~2 downloads per run in sample mode.

cc @felixfelix-bot — please review; team flow is merge on your approval. Suggest landing this and #405 before the alpha2/final tag so the canary publish runs through the verified path.


Amperstrand:
**Branch rebuilt off current `main`** ([7881e75](https://github.com/OpenTollGate/tollgate-module-basic-go/commit/7881e75)) — clears the conflict and drops the stale baggage shared with #405's old head (backup artifacts, a 13.7 MB committed binary, already-merged #358/#378/#383 changes, and the `DEPENDS` hunks that are #407's scope). The PR is now exactly: `scripts/verify_publication.sh` (unchanged from the reviewed version), the `verify-publication` job + `trigger-build-os` dependency + `BLOSSOM_MIN_SUCCESS` tag policy in `build-package.yml`, and a changelog entry. YAML parses, `bash -n` + arg-guard smoke pass, `define-package-matrix.outputs.matrix` verified to exist on main. The functional evidence in the PR body (live v0.5.0/alpha1 runs) still stands since the script is byte-identical. felix review still needed for 0.6.0.

c03rad0r (approved):

PR #113, opened 2026-05-04
feat(cli): JSON output and config/health subcommands

Amperstrand:
## Summary

Adds machine-readable JSON output to the `tollgate` CLI client, config get/set/schema/health handlers over the Unix socket, and security/quality improvements to the HTTP server.

Depends on #112 (schema + dot-path system).

## Changes

**New files:**
- `src/cli/config.go` — 6 socket handlers: `config get`, `config set`, `config schema`, `config save`, `config save-identities`, and config routing
- `src/cmd/tollgate-cli/contract_test.go` — JSON shape contracts for all CLI response types (304 lines)

**Modified files:**
- `src/cmd/tollgate-cli/main.go` — Persistent `--json` flag, `sendCommandRaw`, `printJSON`, config/health cobra subcommands
- `src/cli/server.go` — Added `config` and `health` command routing, `handleHealthCommand`
- `src/main.go` — `getMacAddress` rewritten in pure Go (was `exec.Command("sh", "-c", ...)`), CORS restricted to local origins (was `*`), `log.SetFlags(0)` + `log.SetPrefix("")` + `log.SetOutput(os.Stdout)` to fix double timestamps in syslog

## Design decisions

- **CLI socket protocol unchanged** — config commands use the same `CLIMessage`/`CLIResponse` JSON protocol over `/var/run/tollgate.sock`
- **`--json` flag is persistent** — works with any command: `tollgate --json wallet balance`, `tollgate --json health`
- **No new dependencies** — no logrus migration, no session persistence. Those will be separate PRs.
- **CORS now local-only** — rejects origins from public IPs. The LuCI admin UI is served from the router itself.

## Test plan

- [x] `go test ./src/cmd/tollgate-cli/` — contract tests pass
- [x] `go build ./src/` — main module compiles
- [x] On-device: `tollgate --json config schema | jq '.[0].json_key'` returns `config_version`
- [x] On-device: `tollgate --json health | jq '.success'` returns `true`
- [x] On-device: 58/58 Playwright tests pass via `tollgate --json` CLI backend
- [x] CI: all compile + package jobs pass (publish-metadata fails — expected, fork lacks Blossom secrets)

PR #93, opened 2026-04-21
Add lightning support, balance page, and local SDK packaging helper

alexxie16:
## Summary
- add Lightning invoice generation/status handling plus the backend updates needed to grant access and expose session balance data
- update the captive portal bundle with Lightning UI, a balance landing page, and the supporting localized copy/assets
- add and rename the local OpenWrt SDK packaging helper, preserve apk/ipk local builds, and clean up stale packaging/debug paths while persisting Lightning quote state across restarts

## Test Plan
- `go test github.com/OpenTollGate/tollgate-module-basic-go/src/merchant`
- `go test ./...` (from `src/`)

PR #407, opened 2026-09-18
fix(packaging): drop jq from maintainer scripts; Go owns install_time

Amperstrand:
## Why

Found during QEMU v0.5.0→v0.6.0 upgrade emulation (PRTA campaign #113; bench in physical-router-test-automation@40d1b56). **Upgrades abort on jq-less systems**, and the aborted opkg transaction then orphan-removes runtime deps (nodogsplash…), breaking the portal: Amperstrand/tollgate-module-basic-go#93 (repro + env forensics there).

jq is no longer a package dependency (`DEPENDS:=+libc`), so a bare `jq` in maintainer scripts fails on minimal installs — exactly the offline/no-internet devices the update path must serve.

## What

1. **`packaging/preinst` + the SDK Makefile `preinst` define are no-ops.** The install_time bump they attempted is (a) cosmetic, (b) redundant — the binary owns install.json.
2. **`EnsureDefaultInstall` stamps `install_time` when zero** (fresh installs previously wrote 0) and is now the file's sole writer. Existing timestamps are preserved byte-for-byte — no rewrite on ordinary loads (verified by test, and by sha256-preservation evidence on the upgrade bench).
3. **`check_package_path` rewritten without jq** — targeted sed on the `package_path` string field, safe because install.json is exclusively Go-written (MarshalIndent, stable order). Also fixes a latent bug: it read `.package_path` but wrote `.update_path = null`, so the pending-update path never actually cleared.

## Verification

- `go test ./src/config_manager/...` green (3 new tests: stamps-zero-on-create, preserves-existing-timestamp + no-rewrite, stamps-zero-on-load).
- shellcheck clean.
- The bench evidence behind the design: config/wallet/identities sha-preservation across upgrade→rollback→upgrade with a funded wallet (PRTA #106 comment).

## Notes for reviewers

- Behavior change: install_time now reflects **first service start** (Go) rather than package-unpack time (preinst). For fresh installs these coincide within seconds; for upgrades the existing timestamp is kept as-is.
- `files/usr/bin/check_package_path` field semantics unchanged (`"false"` = no pending update); only the clearing bug and the jq dependency are fixed.


Amperstrand:
# Review — clean and focused; land after approval

Reviewed the diff (packaging + config_manager install path; unit tests only, no router-visible surface claimed). What it does:

- `preinst` becomes a no-op with the jq block removed and a historical note kept.
- `check_package_path` switches jq → sed and fixes a real `.package_path`/`.update_path` read/write mismatch on the way, plus tightens the disabled-value check.
- `install_time` stamping moves to Go (`config_manager_install.go`) with sole-writer semantics — stamp on create / invalid / zero-load, byte-for-byte no-rewrite otherwise — covered by three new tests including the no-rewrite case.

File set is disjoint from the other v0.6.0 PRs — no conflicts, order-independent. One optional nit: call out the `.package_path`/`.update_path` fix in the squash-merge message; it's a behavior change worth grepping later.

Amperstrand:
Second-agent verification pass:

- Root cause is real and matches the fork forensics (#93): `DEPENDS` is `libc` only (verified in the packaging pipeline's own build-ipk invocation), so a bare `jq` in maintainer scripts fails on minimal systems — and the aborted opkg transaction orphan-removing runtime deps is exactly the failure mode worth killing.
- The sed-for-jq substitution in `check_package_path` is sound under the stated invariant: `install.json` is written solely by the Go side via `MarshalIndent` (stable field order), the sed targets one string field, and the `"false"` sentinel matches the Go default.
- Go-side ownership confirmed on main: `config_manager_install.go` carries `EnsureDefaultInstall` with `InstallTimestamp` stamping, so the no-op preinst doesn't lose the install_time behavior.
- The explicit no-op preinst with the historical note is the right call — keeps the packaging contract visible.

Verified from the diff's logic. Worth a router-shaped confirmation (QEMU lane or physical spot-check) that an upgrade over a pre-#407 install cleans up cleanly, since maintainer-script changes only prove out in situ.

Amperstrand:
**Review follow-ups landed — branch rewritten to `0fcabf9` (rebased), MERGEABLE against current `main`:**

| Review ask | Done | Evidence |
|---|---|---|
| Rebase so the ipk test picks up #405's tar-format pin | Branch rebased onto `main` @ `125f6a0` (was 10 commits behind, base `d0cb92f2`) | `git diff origin/main --stat` now shows exactly the PR's 5 files (+125/−49); `packaging/build-ipk.sh:92,97` carry `--format=gnu`; `gh pr view 407` → `mergeable=MERGEABLE` |
| Add the missing CHANGELOG entry (user-visible on the upgrade path) | `Fixed` entry added under `[Unreleased]`: the preinst abort on jq-less devices (which orphan-removes runtime deps), Go ownership of `install_time`, fresh installs untouched, PR link in the house style | `0fcabf9 docs(changelog): add the missing [#407] entry` |

Verification on the rebased head: `gofmt -l` clean, `go vet ./...` and `go build ./...` OK in `src/config_manager`, and `go test -race -count=1 ./...` → `ok …src/config_manager 1.074s` (includes this PR's `config_manager_install_test.go`). Packaging scripts unchanged by the rebase (no conflict).

One note for the re-review: the branch lives on the `Amperstrand` fork and was force-pushed (rebase), so the old head `11a6dec` is orphaned — reviews should reference `0fcabf9` / `058c39b`.

cc @felixfelix-bot — both asks addressed, ready for another pass.

Amperstrand (commented):
Ran `src/config_manager` on the PR head (`refs/pull/407/head`): `go vet` + `go test -race -count=1 -tags testenv ./...` green, including the three new `TestEnsureDefaultInstall*` tests (stamp-on-fresh, preserve-existing + no-rewrite-on-noop, stamp-zero-in-loaded-file). The preinst no-op is sound: the forensics in Amperstrand/tollgate-module-basic-go#93 (jq-less minimal systems aborting the upgrade, orphan-removing runtime deps) justify moving install_time ownership to the binary, and EnsureDefaultInstall's stamp-on-zero covers fresh installs and the zero-valued files the old preinst's failure path could leave. check_package_path's sed is scoped to the single string field the Go side writes with stable ordering, and the `"false"` sentinel handling matches the documented Go default.

Amperstrand:
**Verification pass — behavior validated end-to-end, two non-blocking items flagged.**

Validated:

- **Module battery** on the branch: `gofmt`/`go vet` clean, `go test -race -count=1 -tags testenv ./...` green in `src/config_manager` (incl. the new `config_manager_install_test.go` covering the zero-stamp paths).
- **The sed-based `check_package_path`, live in a sandbox** (mock `apk`, no jq on PATH): staged `package_path` + existing file → install fires → field round-trips to `"false"` via the targeted sed, JSON otherwise byte-identical; `package_path: "false"` → clean no-op, zero install attempts. No jq anywhere on the path.
- **No-op preinst**: exits 0 trivially; the historical-note comment block is accurate (cross-checked against Amperstrand/tollgate-module-basic-go#93 — the jq-less abort + orphan-removal cascade).

Flagged (non-blocking but should be addressed before merge):

1. **Missing CHANGELOG entry.** Per AGENTS.md every user-visible change lands with one, and this changes upgrade behavior on jq-less minimal systems (fixes aborted upgrades). One bullet under `Fixed` linking this PR.
2. **Base freshness:** 11 commits behind `main`. No textual conflict expected (only `packaging/build-ipk.sh` moved on main, untouched here) but given the #413 lesson this session — that PR's own test failed on its stale base and passed after rebase — a rebase before merge is cheap insurance.

Also worth stating for the release thread: the change is regression-safe for the standard flow — `EnsureDefaultInstall` stamps `install_time` on zero values (unit-tested), so first-boot and `preinst`-less upgrades produce the same observable state as before, minus the jq dependency.

c03rad0r (approved):

Issue #403, opened 2026-09-17
Payment consumed with no session when gate-open fails after successful Receive (no refund path)


Amperstrand:
**Fund-safety edge (same family as #375):** a successful payment is consumed even when opening the gate subsequently fails — the customer's ecash is swapped into the operator wallet, no session is granted, and there is no refund path.

**Code path (verified on `main` @ `373770a`):** `PurchaseSession` (`src/merchant/merchant.go:543–560`) calls `Receive` (proofs swapped — irreversible once the mint accepts), then `grantSessionAccess` → `openGateForSession` → `valve.OpenGateUntil` → `ndsctl auth`. On failure, `restoreSession` (`merchant.go:1065`) rewinds **only the in-memory `customerSessions` map** — the customer's value stays in the operator wallet. The caller receives a kind-21023 `session-error` notice; nothing returns or credits the paid amount.

**Live evidence (QEMU OpenWrt 24.10 lab, 2026-09-17, testnut):**
```
12:03:44 PurchaseSession: calling Receive for mint=https://testnut.cashu.exchange token_amount=4 mac=02:11:22:33:44:51
12:03:46 PurchaseSession: Receive completed, amount=3, err=<nil>        # 4 sat − 1 swap fee, now in operator wallet
12:03:46 Opening gate until timestamp duration_seconds=90 mac_address=02:11:22:33:44:51
12:03:47 Error authorizing MAC address error="exit status 1"            # ndsctl auth failed
```
HTTP response: `400` kind-21023, `code=session-error`, `"failed to open gate: error authorizing MAC: exit status 1"`. Reproduced twice (a second occurrence during soak bring-up at 12:06:30Z). Wallet balance grew by the received amount both times.

**Trigger surface:** any gate-open failure after a successful `Receive`. Observed triggers: (a) client MAC not yet registered in NDS at payment time (ndsctl auth exits 1 for unknown MACs), (b) NDS 5.0.2's `ndsctl auth` on an *already-Authenticated* client also exits 1 — first payment after fresh daemon state with a still-authed client hits this (noted on OpenTollGate/physical-router-test-automation#110).

**Impact:** worthless on signet/fakewallet, but on mainnet a paying customer whose gate-open fails loses the payment with no session and no recourse.

**Suggested fix shapes (maintainer decision needed — not unilaterally chosen):**
[...]

PR #412, opened 2026-09-19
fix(merchant): refuse payment when NDS has no client session for the MAC (#403 L1)

Amperstrand:
L1 of #403 — refuse payment up front when the gate provably cannot open, and stop failing auth for already-Authenticated clients.

## Problem

`PurchaseSession` consumes the customer's token (`Receive` swaps the proofs — irreversible) **before** attempting gate-open. When `ndsctl auth` then fails, `restoreSession` rewinds only the in-memory session map: the customer's value stays in the operator wallet, no session, no refund path. Both triggers reproduced in the QEMU OpenWrt lab:

- **(a)** client MAC not yet registered in NDS — `ndsctl auth` exits 1 for unknown MACs;
- **(b)** NDS 5.0.2 `ndsctl auth` on an *already-Authenticated* client also exits 1 — the first payment after fresh daemon state with a still-authed client.

The HTTP handler's fallback MAC `00:00:00:00:00:00` (used when MAC lookup fails, `main.go`) is an instance of (a) that can never be authorized.

## Change

- **`valve.CheckClientState(mac)`** — new read-only probe over `ndsctl json <mac>`: distinguishes *definitively not registered* (empty `{}`) from *registered (+ `Authenticated` state)*, with probe failures surfacing as errors. No state is changed, nothing is authorized.
- **`authorizeMAC`** — on a failed auth attempt, probes once: if the client is already `Authenticated`, the gate is open and the failure is treated as success (trigger (b)). Happy path is unchanged — zero extra ndsctl calls when auth succeeds first try.
- **`PurchaseSession` pre-flight** — before the `Receive` goroutine, re-probes registration with the same cadence/budget as the valve auth retry (5 × 400 ms) so the two-router reseller flow's asynchronous NDS registration is not refused on first sight; if still unregistered, refuses the payment with a kind-21023 `client-not-registered` notice ("Reconnect to the TollGate Wi-Fi and try again") while the customer still holds the token. **Probe errors fail open** — a broken probe must not become a payment denial of service.

Why open-the-gate-first is not the shape: auth is reversible, settlement is not; granting before payment opens a scriptable free-internet window (MAC randomization) bounded only by the mint roundtrip.

## What this does NOT cover (tracked on #403)

The check→auth race (client vanishes mid-transaction) and a fully-dead NDS still leave a tiny post-`Receive` failure window. Closing it needs the L2 (stop discarding the allotment; retry with deadline) / L3 (refund-in-notice) decision — product semantics, maintainer call.

## Verification

- `gofmt -l .` clean, `go vet ./...` and `go build ./...` clean in the root, `merchant`, and `valve` modules.
- `go test -race -count=1 -tags testenv ./...`: root ok, `merchant` ok (202 s), `valve` ok.
- New tests: pre-flight refuses unregistered clients without calling `Receive`; probe errors fail open; the probe retries until registration; `authorizeMAC` succeeds for already-Authenticated clients and keeps retrying for merely-registered ones. The two pre-existing authorizeMAC retry tests were adjusted to count only `auth` invocations (the probe adds `json` calls on the failure path).
- **Not yet validated on a router** — unit-level only; the QEMU lab from #403 can regression-test both triggers.

Refs #403 (L1 scope; L2/L3 remain).


Amperstrand:
# Verification + review notes — good to land after the QEMU regression and #409 co-landing

Verified on head `e0aed2a8` in a clean worktree: `gofmt -l .`, `go vet ./...`, `go build ./...` all exit 0; `go test -race -count=1 -tags testenv ./...` exit 0 (~24 s); and the nested modules the documented gate does not cover: `src/merchant` exit 0 (~215 s), `src/valve` exit 0.

The diff walks the L1 design from #403 exactly: `valve.CheckClientState` (read-only `ndsctl json`, empty `{}` ⇒ not registered), `PurchaseSession` refuses **before** `Receive` after a 5×400 ms re-probe (so the two-router reseller flow's asynchronous NDS registration isn't refused on first sight), probe errors fail open (a broken probe must not become a payment DoS), the `00:00:00:00:00:00` fallback MAC can no longer reach the consume-then-fail path, and `authorizeMAC` short-circuits when the probe reports `Authenticated` — defusing trigger (b) at every call site. New tests pin refusal-before-`Receive`, fail-open, retry-until-registered, and the already-authed path. The two pre-existing valve retry tests now count only `auth` calls and stub `json` with a `Pending` client — a deliberate, stricter semantics; worth a maintainer eyeball since it changes what those tests assert.

Nothing blocking. Notes:
- **Co-landing with #409**: both PRs insert a pre-`Receive` guard at the same spot in `PurchaseSession` — textual conflict, and the intended resolution is to keep both guards (swap-fee check first, then client-not-registered); the stub commit here already anticipates it.
- `mergeable` reads UNKNOWN at review time (GitHub hadn't computed it) — recheck before merge.
- Per the author's own scope note this is L1-only and unit-verified; the remaining pre-merge box is the QEMU regression of both #403 triggers ((a) unknown-MAC refusal, (b) already-Authenticated short-circuit).

Amperstrand:
**QEMU-venue regression of both #403 triggers — PASS.** The unit-verified status in the L1 note is now backed by a live OpenWrt run (24.10.1 x86/64 QEMU venue, real nodogsplash 5.0.2, cdk-mintd FakeWallet mint, this PR's binary deployed on the router):

- **Trigger (a) — refusal before `Receive`:** purchase from an unregistered MAC → HTTP 400, kind 21023, `code=client-not-registered`, in **1.617 s** (the 5×400 ms re-probe profile), router log `not registered in NDS after 5 probes; refusing payment before Receive`, **no `calling Receive` line**, and the refused token was later spent successfully — it was never consumed. (A re-registration race mid-run produced an unplanned second reproduction with identical signature.)
- **Trigger (b) — already-Authenticated short-circuit:** this venue's NDS 5.0.2 does exit 1 on re-auth (characterized first: rc=0 then rc=1 "Probably already authenticated"); with this PR the log shows `Client already Authenticated in NDS; treating gate as open (attempt=1)` and the session was granted.
- **Contrast + happy path:** registered client pays in 56 ms (first-probe preflight hit) vs 1.617 s refusal; full E2E green through real internet egress through the gate.

Note: the PR head is now `9191de7` (rebased since review; same three commits, fix content verified present). With this, the PR's own remaining pre-merge item ("QEMU lab should regression-test both triggers") is satisfied.

c03rad0r (approved):

PR #413, opened 2026-09-19
test(cloud-lab): fee-charging mint + swap-fee e2e suite (#409 follow-up)

Amperstrand:
Fee-charging mint coverage for the cloud lab — the e2e counterpart to #409 that its review asked for ("a cloud-lab contract case with a 1-sat token on a fee-charging mint would pin the classifiers against reality").

Real-world mints charge input fees per swapped proof (`input_fee_ppk`); the lab only had a zero-fee mint, so the swap-fee pre-check path could never fire end-to-end and `GonutsWallet.SwapFeeSats` had zero coverage against a real mint.

## What

- **`mint-fees` service** — second cdk-mintd FakeWallet mint with `CDK_MINTD_INPUT_FEE_PPK=100` (verified knob against cdk-mintd 0.17.6 source: `env_vars/common.rs` → `set_unit_fee`), mirroring mint.coinos.io where the #409 live observation came from: a single-proof swap costs `ceil(100/1000)` = 1 sat. Different mnemonic → different keysets from the free mint.
- **`test_swap_fees.py`** — four tests, run against the real HTTP surface:
  1. the fee is visible in `/v1/keysets` (guards the fixture itself),
  2. a 1-sat token is refused **before the swap** with `payment-error-below-swap-fee`, and refuses *identically twice* — proving the token was never spent,
  3. a 100-sat token (3 proofs → 1 sat fee) is credited **net of fee**: allotment is exactly 99 steps, pinning `GonutsWallet.SwapFeeSats` + `resolveKeysetID` against real cdk-mintd keysets,
  4. zero-fee payments on the free mint are unchanged.
- Upstream lab config accepts the fee mint; client container waits on its healthcheck.

## Two lab traps found while writing this (worth knowing)

- Payment MAC comes from the **`?mac=` query param** (splash-page preauth style), not the event's `device-identifier` tag — without it, every payment falls back to the client IP's lease MAC. Tests now pass per-test MACs via the param (allotments are cumulative per MAC).
- Root-owned `__pycache__` from container runs can serve stale pytest bytecode to later host/container runs; the suite invocation clears it (and `tests/cloud-lab/__pycache__/` is now gitignored).

## Verification (real docker run, this branch + #409 merged)

```
test_swap_fees.py     4 passed
test_smoke_payment.py 6 passed   (no regression)
```

Fee premise verified at the HTTP surface: `mint-fees` keyset reports `input_fee_ppk: 100`, free mint stays `0`.

**Depends on #409** for the refusal path (verified on a merged tree; mergeable in either order — the tests are additive).

Milestone: v0.6.0 (with #409).


Amperstrand:
**Review follow-ups landed — branch rebased onto `main` @ `125f6a0` and rewritten to `2af6f4a`** (the branch previously forked at `a6e01bd`, i.e. **before** #409, so the suite was never exercising the real pre-check on its own base):

| Finding | Fix | Evidence |
|---|---|---|
| Suite cannot tell the pre-check from the post-swap classifier | The below-fee test now asserts the **pre-check's notice content** — `"charges a 1 sat swap fee, so there is nothing left to spend"` — which the classifier's wording (`"charges a swap fee that leaves nothing left to spend"`) does not match | **Mutation-verified on the live lab**: deleted `merchant.go`'s pre-flight block → rebuilt upstream → test goes red (`payment-processing-failed` ≠ expected content/code path); restored → green again |
| "Token stays unspent" never actually queried | New NUT-07 oracle: pure-python `hash_to_curve` (byte-compatible with gonuts' `crypto.HashToCurve`, `Secp256k1_HashToCurve_Cashu_` domain, le32 counter, even-Y) derives each proof's Y; the test POSTs `/v1/checkstate` and requires `UNSPENT` after each refusal | Live run: `POST /v1/checkstate` → `200 {"states":[{"state":"UNSPENT",...}]}`; oracle cross-checked against gonuts v0.11.2 source |
| CHANGELOG conflict | Resolved in the rebase (both #405's and this PR's entries kept; PR link already corrected to #413 by `d76a856`) | `git merge-tree` clean vs `main`; `mergeable=MERGEABLE` |

**Suite result on the rebased head, against the live cloud lab** (upstream image rebuilt from this branch):

```
test_swap_fees.py::TestSwapFees::test_fee_mint_charges_input_fee PASSED
test_swap_fees.py::TestSwapFees::test_below_swap_fee_refused_before_spend PASSED
test_swap_fees.py::TestSwapFees::test_above_swap_fee_succeeds_with_fee_deducted PASSED
test_swap_fees.py::TestSwapFees::test_fee_mint_payments_do_not_break_the_free_mint_path PASSED
4 passed in 5.27s
```

One robustness fix folded in (observed live during validation): test MACs are now randomized per session (locally-administered `02:…`) — the lab is a shared compose project, and two concurrent suite runs on the same hardcoded MACs double each other's cumulative allotments, which is exactly how a phantom failure appeared mid-validation.

Note for the re-review: branch was force-pushed (rebase), so reference `2af6f4a`.

cc @felixfelix-bot — both findings addressed with live evidence, ready for another pass.

Amperstrand:
**Fresh verification on current `main` — the branch needs a rebase before merge; after rebasing locally the lane is 4/4.**

**Ran as-is (pristine `pr/413` @ 4e38d60):** cloud-lab fee-mint lane, fresh `--build`, host-port override in place (the #436 gotchas honored):

```
test_swap_fees.py: 3 passed, 1 failed
FAILED test_below_swap_fee_refused_before_spend
  Expected payment-error-below-swap-fee, got: payment-processing-failed
upstream log: Receive err = could not swap proofs: Inputs: 0, Outputs: 0,
  expected_fee: 0. Transaction inputs should equal outputs less fee
```

**Root cause — base freshness, not the lane.** `pr/413` forks from `a6e01bd` (#408, Sep 18), i.e. **before** #409 (`040dd7f`) landed `isBelowSwapFeeError` + the `payment-error-below-swap-fee` branch in `src/merchant/merchant.go`. The PR's own test expects a notice code its tree cannot produce. (Any earlier "4/4" run must have had #409 in the tree via checkout ordering.)

**Proof of fix:** rebased `pr/413` onto current `main` locally (`verify/413-rebased`, 5 commits replayed; only CHANGELOG needed rerere) and reran the identical lane — **4 passed in 5.17s, exit 0**, including the below-fee refusal with the token left unspent.

Ask: rebase onto `main` (or merge with the knowledge that the branch content itself is verified green post-rebase). No code changes requested — the tests, compose service (`mint-fees` @ `CDK_MINTD_INPUT_FEE_PPK=100`, pinned `cdk-mintd 0.17.6 --locked`), and per-test MAC attribution via `?mac=` all check out on read.

Amperstrand (commented):
Technical review (author's account — cannot approve): the fee-charging mint lane pins the #409 semantics end-to-end (fee-charging NUT-01 test mint + swap-fee e2e); structure matches the existing cloud-lab lanes. Not lab-run by me — needs an approving review and a lab pass.

Amperstrand:
Follow-up at head `2af6f4a` → `8c202b7` (rebased onto main@`e2e3df9`), on the live review:

**F1 — the counter already starts at 0.** The oracle's loop is `range(1 << 16)` — counter 0 first, same as gonuts (`bdhke.go:35`). Verified byte-exact against gonuts' own `TestHashToCurve` vectors. The substantive half of F1 was real though — **no negative control** — and is now closed two ways:

- `test_hash_to_curve_matches_gonuts_vectors` pins the pure-python `hash_to_curve` to the `crypto/bdhke_test.go` vectors. The first vector resolves at counter **0**, so an off-by-one in the loop start (the exact drift F1 feared) fails the suite instantly — checked by mutation: starting the loop at 1 no longer produces `024cce9…`.
- Negative control: the above-fee test now spends its token and requires `assert_token_unspent` to **fail with SPENT** — the oracle is proven able to detect spentness against the real mint, so the UNSPENT passes in the refusal test carry meaning.

**F3 — done:** the three randomized session MACs are printed (autouse session fixture; shows up in any failing test's setup capture, live under `-s`).

**F2 — agreed, follow-up:** a distinct sub-code on the pre-check would pin the path durably instead of message prose, but that is a `merchant.go` change and out of scope for this test/lab-only PR.

Lab (real compose run, rebased head): `test_swap_fees.py` **5/5**, `test_smoke_payment.py` **6/6**. Go gates on the rebased tree: gofmt clean, `go vet`, `go build`, `go test -race -count=1 -tags testenv ./...` all green.

Rebase note: the predicted one-hunk CHANGELOG conflict never materialized — the 3-way merge absorbed it.

Amperstrand:
**F1/F3 verification on `8c202b7` — independently re-run on the live cloud lab:**

```
test_swap_fees.py::TestSwapFees::test_hash_to_curve_matches_gonuts_vectors PASSED
test_swap_fees.py::TestSwapFees::test_fee_mint_charges_input_fee PASSED
test_swap_fees.py::TestSwapFees::test_below_swap_fee_refused_before_spend PASSED
test_swap_fees.py::TestSwapFees::test_above_swap_fee_succeeds_with_fee_deducted PASSED
test_swap_fees.py::TestSwapFees::test_fee_mint_payments_do_not_break_the_free_mint_path PASSED
5 passed in 8.26s
```

How this closes the re-review's asks:

- **F1 (vacuous oracle)** — closed two ways in `8c202b7`: the gonuts-vector pin (the first vector resolves at counter=0, so an off-by-one loop start fails *there*, before any checkstate call) and the spent-proof negative control inside the above-fee test (`pytest.raises(..., match="SPENT")` on the token that payment just consumed). I additionally mutation-tested the exact off-by-one the review hypothesized — starting the counter at 1 — against a lab run: the negative control goes **red** while the unspent assertions stay green, which is precisely the vacuity split the control exists to catch. For the record, the committed code's `range(1 << 16)` starts at 0, matching gonuts' `var counter uint32 = 0`; but as the review correctly argued, only the negative control could have *shown* that.
- **F3 (replayable MACs)** — closed by the autouse session fixture printing the three randomized MACs (visible in the run output above).
- **Rebase note** — branch is on `main` @ `e2e3df9`, mergeable, no CHANGELOG conflict (the dedup commits absorbed what would have collided with #435's hunk).

Process disclosure: a parallel session and I both implemented F1 within the same minute — my variant (dedicated spent-token test + prose counter-parity note) was briefly force-pushed over theirs before I noticed; the branch was restored to their `8c202b7`, whose vector-pinning approach is the stronger of the two. No work was lost; the timeline above is the surviving head.

cc @felixfelix-bot — F1 (with vector pinning + negative control) and F3 are in at `8c202b7`, verified live; F2 (sub-code instead of prose pinning) noted as the follow-up the review itself deferred.

Issue #415, opened 2026-09-19
Test lane: fee-charging signet mint on CLN (inr2) + rune-authorized driver


Amperstrand:
**Test-lane request: fee-charging signet mint backed by our CLN (inr2.cashu.exchange), rune-authorized test driver — the "real world" tier of the fee-mirroring test programme.**

**Why:** real-world mints charge input fees; our signet mint does not. Verified today: `testnut.cashu.exchange` is a nutshell FakeWallet mint whose keysets all report `input_fee_ppk: 0` — every fee behavior #409/#413 hardened (below-fee refusal, net-of-fee credit, keyset resolution) is currently invisible on the fleet that most resembles production. The lab's `mint-fees` (#413) covers the hermetic tier; what's missing is the signet tier where **settlement is real** (melt → Lightning invoice → payout lands) and keyset IDs/rotation shapes come from a real operator setup.

**Proposed architecture (two components):**

1. **`mint-signet-fees`** — cdk-mintd 0.17.6 with the CLN backend (env vars verified against the 0.17.6 source, `src/env_vars/cln.rs`): `CDK_MINTD_LN_BACKEND=cln`, `CDK_MINTD_CLN_RPC_PATH=<lightning-rpc socket>`, `CDK_MINTD_CLN_FEE_PERCENT`, `CDK_MINTD_CLN_RESERVE_FEE_MIN`, plus `CDK_MINTD_INPUT_FEE_PPK=100` to mirror coinos. cdk's CLN backend speaks the **raw UNIX socket**, not clnrest — so the wiring depends on what inr2 exposes (see decision below).
2. **Rune-authorized test driver** — the test runner gets a CLN **rune** (restricted: `invoice`, `listpays`, `listfunds`) and drives/asserts the settlement loop directly: melt with fees → invoice created via rune → signet payment → payout observed. Runes also gate the driver's access without handing it the node.

**Decision needed (infra, @Amperstrand):** does inr2 expose `lightning-rpc` co-locatable (mount/socket-forward) or only clnrest (HTTPS+rune)?
- **(a) socket available** → run `mint-signet-fees` co-located with (or SSH-forwarded to) the socket; no rune needed for the mint itself; driver uses clnrest+rune.
- **(b) clnrest only** → a small socket↔clnrest bridge is required for cdk-mintd, or we contribute/reuse an existing one; alternatively evaluate cdk's `grpc-processor`/LND backends if the node multiplexes those.

[...]

Issue #95, opened 2026-04-22
Self signed certificates / .local


c03rad0r:
Use self signed certificates and/or .local to make the camera work in the captive portal for scanning QR codes

Amperstrand:
The LuCI admin UI (PR #110) runs over HTTP since uhttpd serves it. The captive portal's QR scanner requires HTTPS / camera permission which the WebView won't grant over plain HTTP. The LuCI admin UI itself doesn't need camera access, so this is only a captive portal concern.

If self-signed certs or `.local` mDNS are configured in the future, the LuCI admin UI's WiFi tab could show a certificate status indicator and the captive portal could gain camera access for QR scanning.

c03rad0r:
**Infrastructure shipped on main** by #123 / #142 (Go SSL management rewrite): `tollgate ssl apply` generates a self-signed certificate for the router hostname, configures `gatewaydomainname` (TollGate.lan) and port 443, so the captive portal is served over HTTPS — the secure context a real browser needs to permit camera access (getUserMedia) for QR scanning. (Android's captive-portal WebView still blocks the camera by design, so the portal must be opened in a real browser — which now has a secure context.)

**Pending hardware confirmation** that camera/QR scanning works in a real browser over the self-signed-cert HTTPS endpoint.

**Regression coverage:** https://github.com/OpenTollGate/physical-router-test-automation/pull/36 adds `test_self_signed_cert_serves_portal_over_https_for_qr_scanner` (applies the self-signed cert, asserts the HTTPS portal is reachable with the QR-scanner element present).

Leaving open until the hardware camera/QR check confirms; will close then.

Amperstrand:
## Architecture finding (no fix proposed — needs maintainer decision)

I read the SSL/HTTPS implementation against this issue. The headline finding
is that the recent HTTPS work (#123, #142, plus commits `e8e252e` "serve HTTPS
when TLS certificate files exist" and `612860a` "SNI-based multi-cert TLS
support") **does not address this issue**. By design, the captive portal
splash page is still served over plain HTTP. The HTTPS work targets the LuCI
admin interface only.

### Evidence: the SSL setup explicitly excludes the captive portal

`src/cmd/tollgate-cli/ssl.go` is the user-facing SSL setup tool
(`tollgate ssl apply` / `ssl remove` / `ssl status`). Its own user-facing
output:

```
The captive portal will continue using HTTP interception.
[...]

PR #416, opened 2026-09-19
test(cloud-lab): keyset-rotation lane + swap-fee boundary pins (#409/#413 follow-up)

Amperstrand:
Keyset-rotation lane for the cloud lab + swap-fee boundary pins — the second slice of the fee-mirroring programme (#415 is the signet tier).

Stacks on **#413** (branch includes its commits) and requires **#409** for the classified-refusal path. All verification below ran on a combined tree (#409 + this branch).

## What

- **`mint-rotate` service** — cdk-mintd with `CDK_MINTD_FAKE_WALLET_KEYSET_ROTATIONS` (env verified against the 0.17.6 source). `./run-keyset-rotation.sh` orchestrates two phases from the host: phase A boots one active keyset @100 ppk and the client mints + stashes proofs on it (bind-mounted state file); phase B recreates the container with that keyset **expired with the same mnemonic-derived ID** plus a fresh zero-fee keyset, and the client pays with the stashed proofs.
- **Sum-then-ceil boundary pins** — parametrized 255/1023/2047-sat payments on `mint-fees` decode the actual token, count its proofs, and assert the credited allotment equals `amount − ceil(proofs × 100/1000)` exactly, independent of the wallet's decomposition.

## Two findings surfaced (both pinned in tests, both filed for follow-up)

1. **cdk-mintd 0.17.6 refuses swaps on expired keysets outright** (`"Keyset has expired"`) — pre-rotation proofs become mint-side unspendable. The inactive-but-swappable fee-resolution branch (`mintKeysetFees`) therefore cannot be exercised against cdk at all; that coverage belongs to the signet lane (#415) where a nutshell-style mint keeps rotated keysets swappable.
2. **On a #409 tree, that error classifies as `payment-error-mint-unreachable`** — `isMintUnreachableError`'s `keyset` + `"could not"` pattern matches "could not swap proofs: Keyset has expired", and the branch replaces the message, so the customer is told the mint is temporarily unavailable for a permanently dead token. The test pins the current behavior with a note; details posted on #409.

## Verification (real docker run, combined tree #409+this)

```
run-keyset-rotation.sh: phase A 1/1, phase B 1/1, boundaries 3/3 — all PASSED
test_swap_fees.py + test_smoke_payment.py: 10 passed
teardown: 0 containers left
```

Milestone: v0.6.x (test-only; validates #409 behaviors but does not gate the tag).


Amperstrand:
**Settlement-probe hardening + rebase.** The rotation lane now probes the real quote→PAID path before each phase (`probe_mint_settlement.py`, stdlib, host-side; exit 1 = wedge, 2 = unreachable, with a connect-retry window so a freshly recreated container flapping its port isn't misread) — the /v1/keys healthcheck can't distinguish a wedged cdk-mintd that accepts quotes but never settles (PRTA #95), and this lane's stop/restart rotation is exactly that churn pattern. Branch rebased onto current main (post-#409): the expired-keyset classification pin now runs against the real classifiers. Verified live: both phases probe OK, phase A 1/1 + phase B 1/1 + boundaries 3/3 green, negative control exits 2, teardown clean. Merge order note: #413 then this (still stacked on its commits).

Amperstrand:
**Review follow-ups landed — branch rebuilt as `52012b6`, now stacked cleanly on #413's updated head (`2af6f4a`)** (the branch previously carried #413's original pre-rebase commits and forked below `main`; it is still the superset the review described — its diff vs `main` is unchanged in shape):

| Finding | Fix | Evidence |
|---|---|---|
| Three documents state the opposite of what the assertions pin | The test docstring, both README rows and the CHANGELOG paragraph now state what phase B actually pins: cdk-mintd refuses the swap outright and the router **misclassifies** it as `payment-error-mint-unreachable` while the mint is healthy and the token is permanently dead (the old text claimed a "clean `payment-processing-failed` notice naming the expired keyset", and the README test table even claimed "inactive-keyset fee resolution" — no fee is paid, the swap never happens) | Assertion unchanged (`codes == {"payment-error-mint-unreachable"}`); live phase-B run still passes against it |
| User-facing misclassification deserves a ticket, not a paragraph claiming correctness | **#440 filed** — dedicated expired-keyset code, why it matters for the alpha RC, and the note that the pinned assertion must move in the same change | https://github.com/OpenTollGate/tollgate-module-basic-go/issues/440 |
| Whole-directory collection hazard on the default `client` command | `TestKeysetRotation` now skips unless `ROTATION_LANE=1` (exported by `run-keyset-rotation.sh` into the client container); `TestSwapFeeBoundaries` still runs everywhere. Previously a plain `docker compose run --rm client` ran phase A against a never-rotated mint — stashing a bearer token into the working tree — then failed phase B | Default run collects: `4 skipped` for the phase tests; lane run: both phases execute |

Also fixed while validating (each hit live): the runner's settlement probe 404'd when invoked from outside `tests/cloud-lab` (relative `$0` doubled after `cd` — now `SCRIPT_DIR` is resolved first); the lab's upstream now sets `TOLLGATE_RATE_LIMIT_RPM=600` because the per-IP 10/min limiter made one client-container IP 429 twelve tests mid-suite; rotation MACs randomized per session (same shared-lab accumulation hazard #413 just fixed, observed live here too).

**Live results on this head** (`tests/cloud-lab/`):
- `./run-keyset-rotation.sh` — phase A ✓, phase B ✓ (misclassification pinned), boundaries `[255/1023/2047]` ✓ → `4 passed, 1 deselected`
- default suite, two-router profile: **`19 passed, 3 skipped, 2 warnings in 11.85s`** — 0 failures
- default suite without the profile: 16 passed, 4 skipped, 2 failed (the two `two_router_autopay` tests — pre-existing lab shape: they need `--profile two-router`, unchanged by this PR)

Sequencing per the review stands: #413 first (it now carries the pre-check pin + NUT-07 oracle), then this branch is a clean 3-commit delta on top.

cc @felixfelix-bot — both findings addressed with live evidence, ready for another pass.

Amperstrand:
**Verification on current `main` — same base-freshness caveat as #413, and the lane itself is green post-rebase.**

- Pristine `pr/416` forks 9 commits behind `main` (pre-#409) and *contains* #413's commits, so it inherits the same stale-base failure mode (its bundled `test_swap_fees.py` expects #409's `payment-error-below-swap-fee` classification). Rebased locally onto current `main` (8 commits replayed, rerere-resolved `CHANGELOG.md` only).
- **Post-rebase, full two-phase runner:** `./run-keyset-rotation.sh` → **4 passed, 1 deselected, exit 0** — phase A (active 100-ppk keyset, proofs minted + stashed), phase B (keyset expired via `MINT_ROTATE_ROTATIONS` recreation; stashed proofs surface as the clean classified error, not a crash), plus the sum-then-ceil boundary pins `[255] [1023] [2047]`.
- **Swap-fee suite on the same rebased tree:** 4/4 (from the #413-thread evidence).
- Operational note for reviewers running this lane by hand: invoking `test_keyset_rotation.py` directly (without the runner) fails both phases with self-explanatory assertions ("did run-keyset-rotation.sh start") — the two-phase state + `/tests/.rotation-state.json` only exist via the runner. That cost me two confusing runs before reading the docstrings; the runner is the only supported entry point, which the PR's README section should perhaps state explicitly.
- The settlement probe guard (`probe_mint_settlement.py` before each phase, the PRTA #95 wedge class) is a sound addition and fired correctly in a run where a stale `tg-mint-rotate` container was present.

Ask: rebase onto `main` (stacks on #413's rebase) — content verified green after.

Amperstrand:
**Coordination note for the maintainer (re: the phase-B pin and #447).**

The deliberate misclassification pin in phase B (`codes == {"payment-error-mint-unreachable"}`) now has its fix: **#447** adds the `payment-error-keyset-expired` classification (classifier + health-tracker exclusion), and I've validated the pair end-to-end — the rotation lane green (4 passed) on a combined tree of #447 + this lane with the pin flipped to the new code. Evidence and the exact one-line flip: [#447 comment](https://github.com/OpenTollGate/tollgate-module-basic-go/pull/447#issuecomment-5752727166).

Suggested sequencing: merge #447 first (or same batch), then flip this pin in the same rebase that picks #447 up — exactly the conscious-update flow the pin's comment describes. Happy to push the one-liner to this branch if preferred.

(Also noted: the branch is actively tended — currently only 1 commit behind `main`; my earlier rebase-verification evidence on this thread predates the new follow-up commits like the unspentness proof, so those still want a lane run from whoever pushed them.)

Amperstrand (commented):
Technical review (author's account — cannot approve): keyset-rotation lane + swap-fee boundary pins as the #409/#413 follow-up; TestKeysetRotation/TestSwapFeeBoundaries/TestSwapFees classes target the rotation window where proofs straddle keysets. Not lab-run by me — needs an approving review and a lab pass.

Amperstrand:
**Re-review follow-ups landed in `30790c7`, restacked on #413's verified head (`8c202b7`):**

| Finding | Fix | Live evidence |
|---|---|---|
| **F2** — boundary tests used deterministic MACs in a file whose own comment documents that pattern flaking | Boundaries now use per-session `_mac(amount & 0xFF)` (parametrized cases stay distinct via the suffix; MAC printed in the failure message) | Re-ran the boundary selection against the **still-up lab** right after a full lane run: **3/3 passed** — the exact re-run scenario where the deterministic `02:00:00:00:hi:lo` MACs accumulated and failed the absolute allotment assertions |
| **F3** — `.rotation-state.json` (live bearer token + expired keyset id) survived aborted runs | Cleanup trap now `rm -f`s the state file **unconditionally** (KEEP preserves the lab for inspection, not the ecash) | After a completed `KEEP=1 ./run-keyset-rotation.sh`: state file absent on host and inside the bind mount — the trap is the only remover, so the abort path is the same code |
| **F1** — pin tracked no classifier fix | Every pin site (assertion comment, test docstring, both README rows, CHANGELOG) now references **#447** — the dedicated expired-keyset code implementing #440 — alongside the #415 signet-lane note | grep: `#447` at 2+2+1 sites across test/README/CHANGELOG |

**Verification on `30790c7`:**
- Full lane via `./run-keyset-rotation.sh`: phase A ✓, phase B ✓ (misclassification pinned), boundaries `[255/1023/2047]` ✓ → `4 passed, 1 deselected`
- Default `docker compose run --rm client -sv test_keyset_rotation.py` (no `ROTATION_LANE`): **3 passed, 2 skipped** — gating intact on the restacked head
- `bash -n` + `shellcheck -S warning` clean on the runner; `py_compile` clean

Lab containers and volumes torn down after verification.

cc @felixfelix-bot — F2/F3 fixed in-tree with live evidence, F1 now tracks #447; ready for another pass (RC disposition per the re-review unchanged: land after the cut, after #413, with the classifier decision made separately in #447).

Issue #400, opened 2026-09-17
Degraded mode: runtime downgrade never auto-recovers (stuck until restart)


Amperstrand:
**Bug:** runtime degraded-mode downgrade never auto-recovers — service stays degraded until manually restarted.

**Impact:** any transient mint outage under running traffic (the exact churn class #374 targeted) leaves tollgate-wrt permanently degraded: payments return `service-unavailable`-class notices indefinitely while the mint is healthy again. Observed live: downgrade at 06:55Z, mint probe `ok=true` again at 07:00Z, still degraded at 09:01Z (2h+) until `/etc/init.d/tollgate-wrt restart`.

**Root cause (code-confirmed on `main` @ `373770a`):** the runtime downgrade path (`main.go:118–128`, `registerReachableSetChangedCallback` → `NewMerchantDegradedFromFull`) registers `deg.OnUpgrade(...)` but nothing ever fires it — `SetOnFirstReachableForDegraded` is only called on the startup path (`merchant.go:85`) and the wallet-init-failure path (`merchant.go:132`). The shared tracker's reachable-set callback (`main.go:107–129`) bails on non-full merchants (`full, ok := current.(*merchant.Merchant); if !ok { return }`), so the recovery signal is dropped on the floor.

**Evidence:** OpenTollGate/physical-router-test-automation#110 (Phase D): 8-minute post-unblock window with `mint probe ok=true` and no upgrade; recovery only via restart. Repro: block mint on-router (`/etc/hosts`), let the 5-min probe downgrade, unblock, observe.

**Suggested fix shape:** register the first-reachable upgrade callback in the runtime downgrade path as well (or make the reachable-set callback handle degraded→full transitions instead of returning).

**Recovery-timing note:** aggressive 15s probing with threshold=1 only starts at boot (`StartProactiveChecks` when `reachableCount==0` at startup); runtime recovery rides the 5-min ticker.

Amperstrand:
Fix landed as #421 (TDD cycle pin: degraded start → mint recovery → proactive check → onUpgrade with a real full merchant; race suite green). Together with #420 (#401), the degraded-mode pair is complete: downgrade fires under traffic, and a runtime downgrade recovers without restart. Remaining venue: the QEMU block/unblock regression in the PRTA #110 rig that caught both live.

PR #110, opened 2026-05-01
Refactor LuCI admin UI to OpenWrt best practices

Amperstrand:
## Summary

Replace the custom CGI REST API (`cgi-bin/tollgate-api`) with standard OpenWrt LuCI patterns (`fs.exec_direct`, `fs.read_direct`, `poll.add`, `ui.showModal`, E() DOM builder) — the same approach used by luci-app-adblock and other mainstream LuCI applications.

### Changes

**Removed**
- `files/www/cgi-bin/tollgate-api` — 319-line CGI shell script with custom JSON parsing, URL decoding, and HTTP response handling

**Added**
- `files/usr/share/rpcd/acl.d/luci-app-tollgate-payments.json` — rpcd ACL granting file read/write and exec permissions for tollgate binaries
- `files/usr/libexec/tollgate-luci-helper` — minimal shell script (~30 lines) for structured socket operations (wallet fund/drain, WiFi management) that require direct Unix socket access

**Rewritten**
- `settings.js` — Full rewrite from CGI `fetch()` calls to LuCI standard modules:
  - `'require fs'` for `fs.exec_direct()` and `fs.read_direct()`
  - `'require poll'` for `poll.add()` auto-refresh (5s interval, pauses when tab hidden)
  - `'require ui'` for `ui.showModal()` confirmation dialogs
  - E() DOM builder throughout (no innerHTML)
  - `load()` pre-fetches balance, version, status on page load
  - Six tabs: Overview, Wallet, Network, Configuration, Logs, Advanced

**Updated**
- `Makefile` — removed CGI install lines, added helper + ACL install lines

### Architecture

```
LuCI browser (settings.js)
  ├── fs.exec_direct('/usr/bin/tollgate', [...])     ← display-only CLI commands
  ├── fs.exec_direct('/usr/libexec/tollgate-luci-helper', [...])  ← socket operations
  ├── fs.read_direct('/etc/tollgate/config.json')     ← config reads
  └── fs.write('/etc/tollgate/...', data)             ← config saves (with backup)
```

## Testing

- [ ] Deploy to router and verify all 6 tabs render correctly
- [ ] Verify wallet fund/drain through helper script
- [ ] Verify WiFi rename/password changes
- [ ] Verify service start/stop/restart
- [ ] Verify auto-refresh on Overview and Logs tabs
- [ ] Verify config JSON editor validate/save
- [ ] Verify rpcd ACL permissions are correct

PR #420, opened 2026-09-19
fix(merchant): fire reachable-set callback from MarkUnreachable (#401)

### LINKED ISSUE #401: mint-health: payment-path MarkUnreachable suppresses the degraded-mode transition
**Bug:** a payment failure during a mint outage suppresses the degraded-mode transition entirely — the health tracker can permanently miss an outage that real user traffic already observed.

**Root cause (code-confirmed on `main` @ `373770a`):** `MintHealthTracker.MarkUnreachable` (mint_health_tracker.go:162) decrements `reachableCount` and clears the mint — but does **not** fire `onReachableSetChanged`. The periodic probe path computes `setChanged := newCount != t.reachableCount` (mint_health_tracker.go:255–277) against that already-zeroed count, so `0 != 0` is false → the downgrade callback never fires.

**Sequence (live-reproduced):** mint blocked on-router → payment attempt fails (`Receive` error → `MarkUnreachable`) → count silently drops to 0 with no callback → subsequent probes fail (`mint probe FAILED` logged every 5 min) but `setChanged` stays false → **no downgrade ever happens**. Only a payment-failure-free probe hit triggers the transition (verified positively in the same session).

**Impact:** with real customer traffic present, the first payment of an outage disables degraded-mode accounting for the rest of the outage. Inconsistent with the no-traffic case, where the transition works.

**Evidence:** OpenTollGate/physical-router-test-automation#110 (Phase D, finding D-1): 6 minutes of failing probes + prior payment failure → no transition; fresh-daemon run without the payment → transition fires on the first failing probe.

**Suggested fix shape:** fire `onReachableSetChanged` from `MarkUnreachable` when the reachable set empties (same as the probe path), or make the probes authoritative by not letting the payment path mutate `reachableCount` out-of-band.

Related: #400 (runtime downgrade has no recovery path).

Amperstrand:
Fixes #401 (the suppression half of the degraded-mode bug pair; the recovery-path gap #400 stays tracked separately).

## The bug

`MarkUnreachable` zeroed the reachable count without firing `onReachableSetChanged`, so the probe path's `setChanged := newCount != t.reachableCount` later compared 0 ≠ 0 — with traffic present during a mint outage, **the degraded-mode transition never fired**. Live-reproduced in PRTA #110 Phase D: 6 minutes of failing probes after one payment failure → no transition; the identical run without a payment transitions on the first failing probe.

## The fix

`MarkUnreachable` now fires the callback whenever a **previously-reachable** mint goes down — collected under the lock, fired outside it with `go cb()`, matching the probe path's existing pattern. Already-unreachable mints stay silent (no spurious set-change events).

## Verification (TDD)

- RED first: `TestMarkUnreachable_FiresSetChangedCallback` failed exactly per the issue (callback not fired, 2s timeout).
- GREEN: fix lands, plus `TestMarkUnreachable_UnknownMint_DoesNotFireSetChanged` pins the no-spurious-fire side.
- Full merchant suite: `gofmt`/`go vet` clean, `go test -race -count=1 -tags testenv ./...` green (199 s).

## Deliberately out of scope

- **#400** (runtime downgrade never auto-recovers — the reachable-set callback in `main.go` bails on non-full merchants, so recovery is dropped). With this fix the *downgrade* now fires reliably under traffic; #400's fix (register/handle the upgrade path for runtime downgrades) is the natural next PR and the pair should land before 0.6.0.
- E2E orchestration: killing a mint mid-suite needs host-side control; the QEMU venue that reproduced both bugs is the right regression surface (PRTA #110), the unit pins cover the mechanism.

Milestone: v0.6.0 (degraded-mode correctness is customer-facing under the release's payment-trust theme).


Amperstrand:
**Cloud-lab A/B evidence for the behavioral delta (docker lab, same scenario on `main` vs this branch's binary built from source).**

**Scenario — mint outage *with* traffic (kill mint, then two payments):**

| | baseline (`main` @ `040dd7f`) | this fix |
|---|---|---|
| payment 1 | `400 payment-error-mint-unreachable` | `400 payment-error-mint-unreachable` |
| upstream log | **0** matches for `reachable set changed` / `firing callback` / `downgrading` | `MarkUnreachable: reachable set changed (mint=http://mint:8085), firing callback` → `All mints unreachable — downgrading to degraded mode` — **triggered by the payment itself, no probe wait** |
| payment 2 (+6s) | still `payment-error-mint-unreachable` (full mode persists; transition suppressed) | `400 service-unavailable` "TollGate is initializing. No reachable mints." (degraded mode serving; ~4s race while the degraded wallet loads) |

That is #401 exactly: on the baseline, the payment that observed the outage silenced the transition for the rest of it; with the fix, the downgrade fires the moment traffic sees the mint go down.

*Methodology note for anyone reproducing: `docker compose run client` silently restarts a stopped mint to satisfy its `depends_on` healthcheck — outage simulation must use `--no-deps` and network-disconnect/reconnect (`--alias mint --ip 172.28.0.2` on reconnect, or DNS stays broken). The A/B driver is `tests/cloud-lab/ab_pay.py` (untracked on the demo branch; will be folded into the degraded e2e lane).*

Companion recovery-side evidence: #421.

c03rad0r (approved):

PR #421, opened 2026-09-19
fix(merchant): wire recovery trigger for runtime downgrades (#400)

### LINKED ISSUE #400: Degraded mode: runtime downgrade never auto-recovers (stuck until restart)

Amperstrand:
Fixes #400 — the recovery half of the degraded-mode bug pair (suppression half: #401, fixed in #420).

## The bug

The startup degraded paths (zero-reachable boot, wallet-init failure) register the tracker's first-reachable callback — the *producer* that performs the degraded→full upgrade. The **runtime** downgrade path in `main.go` only registered the `onUpgrade` *consumer*; nothing ever fired it. A service that downgraded under traffic stayed degraded until manually restarted — live evidence: 2h+ degraded with the mint probe healthy again for most of it (PRTA #110 Phase D).

With #420 landed first, the downgrade itself now fires reliably under traffic — which made this dropped recovery signal the remaining half of the pair.

## The fix

- `MerchantDegraded.WireRecoveryTrigger()` — registers `SetOnFirstReachableForDegraded` → `AttemptUpgrade` on the degraded merchant's own tracker.
- `MerchantDegraded.AttemptUpgrade()` — shuts down the degraded wallet, rebuilds a full merchant via the same `newFullMerchant` the startup paths use, fires `onUpgrade`.
- `main.go`'s runtime-downgrade branch calls `deg.WireRecoveryTrigger()` after `deg.OnUpgrade(...)`.

The startup paths keep their inline closures (unchanged, working code); they can migrate to the shared methods later.

## Verification (TDD)

- RED: `TestDegradedWireRecoveryTrigger_RuntimDowngradeUpgrades` failed to compile (method absent).
- GREEN: full cycle pinned in-package — degraded start against a switchable fake mint, recovery, `runProactiveCheck`, and `onUpgrade` fires with a real `*Merchant` (real gonuts wallet rebuilt against the fake mint — not a stub).
- `gofmt`/`go vet`/`go build` clean in root and merchant; full merchant suite `-race -count=1 -tags testenv` green (218 s).

## Out of scope

QEMU regression of the live sequence (block mint → downgrade → unblock → auto-upgrade) belongs in the PRTA #110 venue that reproduced the bug; the unit cycle pins the mechanism.

Milestone: v0.6.0 — with #420 this completes the degraded-mode pair under the release's payment-trust theme.


Amperstrand:
**Cloud-lab A/B evidence for the recovery delta — the baseline stays degraded forever; the fixed build upgrades and serves payments again.**

**Scenario — probe-only outage (no traffic, so the baseline *can* downgrade), then full mint recovery:**

Setup: fresh upstream boot with mint reachable (full mode) → network-disconnect the mint → first failing probe downgrades **both** builds identically:

```
18:46:41 mint probe FAILED: ... lookup mint ... no such host
WARN  All mints unreachable — downgrading to degraded mode
payment → 400 service-unavailable
```

Then reconnect the mint (verified: `DNS+HTTP OK from upstream`) and wait out the recovery threshold (3 consecutive 5-min probe successes; the aggressive 15s mode is boot-only — #400's timing note):

| | baseline (`main` @ `040dd7f`) | this fix |
|---|---|---|
| probes after recovery | `ok=true` ×4 over 16 min (19:21, 19:26, 19:31, 19:36) | `ok=true` ×3 (17:43, 17:48 …) |
| upgrade log lines | **zero** — no `attempting upgrade`, no `Upgrading from degraded` | `Mint became reachable — attempting upgrade from degraded mode` + `Upgrading from degraded to full merchant after recovery` |
| payment after recovery window | **still `400 service-unavailable`** — stuck until manual restart | **`200 kind=1022`** — full session granted again |

The baseline run matches the live 2h+ incident (#400): healthy probes, no upgrade, payments dead until restart. Same binary lineage as the recovery evidence in #420's comment (combined `test/degraded-ab` tree = `main` + #420 + #421).

One honest observation from the run: the fixed build's downgrade→upgrade cycle end-to-end takes ~15 min because runtime recovery rides the 5-min probe ticker with a 3-consecutive-success threshold. That's pre-existing recovery *timing* (not wiring) — if we ever want faster runtime recovery, that's a separate, deliberate tuning decision (e.g. starting the aggressive probe cycle on runtime downgrades too), not part of this fix.

Amperstrand:
**Lab A/B validation of the runtime-downgrade recovery (#400)** — deterministic three-router experiment: deployed this PR's head (`91129d1f`, static linux/amd64 build) to router **B** (treatment), left router **A** on the identical pre-PR build (control, both were md5 `633be02b…`), flapped the mint under both, and observed **B re-arm and upgrade back to a serving merchant with no restart, while A stayed wedged in `no-reachable-mints` indefinitely with healthy probes** — exactly the #400 symptom.

| | Router B — this PR (`91129d1f`) | Router A — pre-PR build (control) |
|---|---|---|
| Baseline | kind 10021, prices served | kind 10021, prices served |
| Mint down (T0) | 11:46:58 | 11:46:58 |
| Degraded (kind 21023) | 11:50:59 (proactive probe: `connection refused` → downgrade) | ≤11:48:13 (proactive probe → downgrade) |
| Mint back up | 11:52:46 (same pubkey `03d902f3…`) | 11:52:46 |
| Probes after restore | `ok=true` at 11:55:59, 12:00:59, 12:05:59 | `ok=true` at 09:58:04Z, 10:03:04Z (≥3 consecutive) |
| Outcome | **12:06:07 advert back to kind 10021 with price tags — no restart, same PID** | **Still kind 21023 at end of window (+30 min)** — only the set-changed callback fired, which no-ops on a degraded merchant |

Verbatim recovery lines from B (`logread`, daemon PID 9952 throughout — no restart):

```
09:50:59 mint probe FAILED: url=http://10.99.99.2:8085/v1/keysets elapsed=905.573µs error=Get "http://10.99.99.2:8085/v1/keysets": dial tcp 10.99.99.2:8085: connect: connection refused
09:50:59 runProactiveCheck: firing callback (hadReachable=true, setChanged=true)
09:50:59 WARN All mints unreachable — downgrading to degraded mode  module=main
09:55:59 mint probe: url=http://10.99.99.2:8085/v1/keysets status=200 keysets=1 elapsed=1.092485ms ok=true
10:00:59 mint probe: url=http://10.99.99.2:8085/v1/keysets status=200 keysets=1 elapsed=1.085501ms ok=true
10:05:59 mint probe: url=http://10.99.99.2:8085/v1/keysets status=200 keysets=1 elapsed=1.167234ms ok=true
10:05:59 runProactiveCheck: firing callback (hadReachable=true, setChanged=true)
10:05:59 Mint became reachable — attempting upgrade from degraded mode
10:05:59 Upgrading from degraded to full merchant after recovery  module=main
```

Control (A), verbatim — probes healthy, nothing re-arms, advert still degraded at 12:06:18:

```
09:58:04 mint probe: url=http://10.99.99.2:8085/v1/keysets status=200 keysets=1 elapsed=1.791291ms ok=true
10:03:04 mint probe: url=http://10.99.99.2:8085/v1/keysets status=200 keysets=1 elapsed=5.311534ms ok=true
10:03:04 runProactiveCheck: firing callback (hadReachable=true, setChanged=true)
10:03:04 Reachable mint set changed — rebuilding merchant  module=main
{"kind":21023,…,"tags":[["level","warning"],["code","no-reachable-mints"]],"content":"TollGate is initializing. No reachable mints detected. Service will auto-recover.",…}
```

**Post-recovery regression (light):** after the upgrade B's rebuilt merchant re-fetched keysets/pricing from the mint (advert `price_per_step` tags present), the payout routine restarted, and the LAN client (de-authed by B's earlier deploy restart) was re-captured by the captive portal (HTTP 307 → splash, splash serves 200). B processed mint round-trips on the recovered wallet; through-traffic beyond A remains blocked only because control A is intentionally left wedged.

**Rig:** three-QEMU-OpenWrt-router chain (client 10.99.99.100 → B → A → C → host), single configured mint `http://10.99.99.2:8085` (cdk-mintd 0.18.0 FakeWallet on the host); flap = SIGTERM + relaunch of the mint process (same data dir + mnemonic, same pubkey verified). B ran PR head `91129d1f` (CGO_ENABLED=0 static build); A ran the pre-PR build unchanged.

**Caveats:** this validates the *runtime* downgrade→recovery path specifically (the startup-degraded path was already covered). One observation for a possible follow-up: after a *runtime* downgrade the 15s aggressive probe loop is not armed (it only starts for startup-degraded), so recovery latency here was bound by the 5-min proactive cycle × `recoveryThreshold=3` (~13 min after the mint returned, worst case ~15 min). The fix itself works — no restart was needed — but arming the aggressive retry on runtime downgrade too would cut recovery to seconds.

Amperstrand (commented):
Approval package for reviewers (all machine-side work done; one approving review + merge remains):

**What this fixes (issue #400):** a runtime downgrade wires the onUpgrade consumer but never the tracker's first-reachable trigger, so a service that downgrades under traffic stays degraded until manually restarted — live incident: 2h+ degraded with the mint probe healthy again (PRTA #110 Phase D). Adds `MerchantDegraded.WireRecoveryTrigger`/`AttemptUpgrade`, wired in the runtime-downgrade path in main; pinned end-to-end by `TestDegradedWireRecoveryTrigger_RuntimDowngradeUpgrades`.

**Branch state:** rebased onto current main (01908f2) — CHANGELOG conflict resolved keeping all entries; sibling fix #420 (tracker callback, issue #401) is already merged.

**Verification on the exact branch tip (125cfa2):** `gofmt -l .` clean, `go vet ./...` clean, `go build ./...` ok, `go test -race -count=1 -tags testenv ./...` ok (22.9s).

**Do not merge #427 instead** — as rebased it reverts #420 (removes mint_health_tracker.go −20 / tests −48); flagged there. #426 is redundant post-#420.

c03rad0r (approved):

